PR #5, opened 2016-09-21
Add experimental celery tools

bloodearnest:


thomir:
Hi,

This looks great - the talisker.celery entry point will be particularly useful. Thanks!

PR #7, opened 2016-09-22
Use string for the X-Request-Id header

fgallina:
1) This follows PEP 3333: https://www.python.org/dev/peps/pep-3333/#a-note-on-string-types
2) Having the X-Request-Id as bytes causes werkzeug to blow with a TypeError when iterating on available headers at this specific point: https://github.com/pallets/werkzeug/blob/master/werkzeug/datastructures.py#L1349

Local runs are also triggering unrelated errors but I haven't time to check those yet: http://paste.ubuntu.com/23216632/

### DIFF
--- a/talisker/request_id.py
+++ b/talisker/request_id.py
@@ -92,7 +92,6 @@ class RequestIdMiddleware(object):
         self.app = app
         self.header = header
         self.wsgi_header = 'HTTP_' + header.upper().replace('-', '_')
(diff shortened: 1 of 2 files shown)


bloodearnest:
Merged, thanks. I could have sworn I'd already added this fix, but it must have got lost in a branch somewhere

Your local test fails are timezone related - I didn't realise time.mktime expects localtime inputs. Fixing in another branch.

PR #12, opened 2016-10-04
Status endpoint should handle bytestrings

thomir:


bloodearnest:
Have pushed a identical fix for this, but without failing tests :)

PR #14, opened 2016-10-10
Better dummy statsd implementaion

bloodearnest:
This version is more correct, and also provides optional message buffering for using in testing.

### DIFF
--- a/talisker/statsd.py
+++ b/talisker/statsd.py
@@ -28,7 +28,7 @@ from contextlib import contextmanager
 from urllib.parse import urlparse, parse_qs
 
 from statsd import defaults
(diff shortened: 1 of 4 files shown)


talisker/statsd.py
```diff
@@ -24,8 +24,11 @@
 from builtins import *  # noqa
 
 import os
+from contextlib import contextmanager
 from urllib.parse import urlparse, parse_qs
-from statsd import StatsClient, defaults
+
+from statsd import defaults
+from statsd.client import StatsClientBase, PipelineBase
```
1stvamp: I don't think `PipelineBase` is ever used?

PR #15, opened 2016-10-19
Access log now uses talisker format

bloodearnest:


bloodearnest:
@1stvamp can you take a look?

bloodearnest:
A question for me is weather we should enable access logs by default or not..

Current behaviour is to not enable by default, but do when DEVEL is set. This means a slightly awkward set up of having to enable access_logfile=- in prod, but leaves you free to route it elsewhere if you want.

Thoughts?

PR #19, opened 2016-11-04
Make the root logger a StructuredLogger

bloodearnest:
Discovered that tags were not working if you logged directly to the root logger. It's an uncommon case, but it's a wart. So fixed it by replacing the logging module global root object with a StructuredLogger. This is not nice, but it's impossible to do anything else, given logging's design.


bloodearnest:
Decided that this is too much cost for too little gain. Will document the limitation instead.

PR #33, opened 2016-11-25
prometheus_client integration

maxiberta:
Add optional `prometheus_client` integration.

### DIFF
--- a/docs/prometheus.rst
+++ b/docs/prometheus.rst
@@ -1,0 +1,33 @@
+.. highlight:: python
+
+
+==========
+Prometheus
+==========
+
+Talisker provides optional `prometheus_client` integration and configuration.
+
+
+Installation
+------------
+
+The package supports extras args to install `prometheus_client`::
+
+    $ pip install talisker[prometheus]
+
+Configuration
+-------------
+
+`prometheus_client` integration needs no extra configuration if running
+in single process mode.
+
+If the app is running in multiprocess mode (ie. with multiple workers), the
+`prometheus_multiproc_dir` envvar must be set to a writable directory that can be used
+for metrics.
+
+Integration
+-----------
+
+If `prometheus_client` is installed, Talisker will expose metrics collected by the
+app using the default registry. Custom registries are only supported in multiprocess mode.
+Metrics are available at /_status/metrics in Prometheus text format.
--- a/docs/prometheus.rst
+++ b/docs/prometheus.rst
@@ -23,7 +23,8 @@ in single process mode.
 If the app is running in `multiprocess mode <https://github.com/prometheus/client_python#multiprocess-mode-gunicorn>`_
 (ie. with multiple workers), the `prometheus_multiproc_dir` envvar should be set
 to a writable directory that can be used for metrics. If this envvar is not found,
-a temporary directory will be created and used instead.
+a temporary directory will be created and used instead (but note that metrics will be reset
+on restarts).
 
 A default `worker_exit <http://docs.gunicorn.org/en/stable/settings.html#worker-exit>`_
 server hook is automatically set up for cleaning up the shared
--- a/docs/prometheus.rst
+++ b/docs/prometheus.rst
@@ -21,11 +21,15 @@ Configuration
 in single process mode.
 
 If the app is running in `multiprocess mode <https://github.com/prometheus/client_python#multiprocess-mode-gunicorn>`_
-(ie. with multiple workers), the `prometheus_multiproc_dir` envvar must be set
-to a writable directory that can be used for metrics. In this case, a default
-`worker_exit <http://docs.gunicorn.org/en/stable/settings.html#worker-exit>`_ server hook is automatically set up for cleaning up the shared
-prometheus directory. Note that the default `worker_exit` can be overridden
-by user supplied configuration.
+(ie. with multiple workers), the `prometheus_multiproc_dir` envvar should be set
+to a writable directory that can be used for metrics. If this envvar is not found,
+a temporary directory will be created and used instead.
+
+A default `worker_exit <http://docs.gunicorn.org/en/stable/settings.html#worker-exit>`_
+server hook is automatically set up for cleaning up the shared
+prometheus directory as well. Note that this can be overridden
+by user supplied configuration, and in that case the custom `worker_exit` function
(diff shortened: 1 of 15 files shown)


bloodearnest:
Looking good.

I think /_status/test/{statsd,prometheus,sentry} would probably be nicer urls, as discussed in IRC

I'm guessing there isn't any way for us to safely create a prometheus_multiproc_dir automatically if worker processes are >1 ? Is persisting that dir between reboots really crucial? If not, a /var/run or /tmp dir could work, perhaps?

bloodearnest:
Oh, and I really appreciate all the doc fix ups! Thanks

maxiberta:
Reusing the same shared dir is useful to persist metrics across restarts, eg. to prevent counters from resetting to zero. But I guess we could leave that decision up to our users? In that case a good idea would be creating and using a temporary directory in `TaliskerApplication.init()` if workers > 1 but no envvar, instead of logging a warning.

maxiberta:
Oh, and btw, forgot about [`prometheus_client.multiprocess.mark_process_dead()`](https://github.com/prometheus/client_python/blob/master/prometheus_client/multiprocess.py#L89). Should be called on `worker_exit`.

talisker/gunicorn.py
```diff
@@ -47,6 +49,17 @@
 }
 
 
+logger = logging.getLogger(__name__)
```
bloodearnest: This probably needs to be moved back to where it was, or else it will be created at import time, which may be *before* the logging has been configured, meaning the logger will be the wrong type, and it won't support tags on logs.

Note: this is only an issue for talisker's own logs, as talisker shouldn't import any application code until after it has configured logging. But in talisker's own code, we need to delay log creation, as some things will be imported before logging has been configured.

talisker/gunicorn.py
```diff
@@ -170,6 +182,8 @@ def init(self, parser, opts, args):
             'loglevel': 'DEBUG',
         })
 
+        cfg['worker_exit'] = prometheus_multiprocess_worker_exit
```
bloodearnest: I think this should probably only be added to cfg iff (workers >1 and prometheus_multiproc_dir in os.environ and util.pkg_is_installed('prometheus_client'))

tests/test_endpoints.py
```diff
@@ -193,22 +194,58 @@ def app(e, sr):
     assert response.status_code == 500
 
 
-def test_error(client):
-    response = client.get('/_status/error',
+@pytest.mark.parametrize('error_url', (
```
bloodearnest: Oh, I do *like* pytest so very much : )

talisker/gunicorn.py
```diff
@@ -179,6 +193,22 @@ def init(self, parser, opts, args):
...
+
+        # Use pip to find out if prometheus_client is available, as
+        # importing it here would break multiprocess metrics
+        if (util.pkg_is_installed('prometheus-client') and
+                (self.cfg.workers or 1) > 1):
+            if 'prometheus_multiproc_dir' not in os.environ:
+                logger.info('running in multiprocess mode but '
+                            '`prometheus_multiproc_dir` envvar not set')
+                tmpdir = tempfile.mkdtemp()
+                os.environ['prometheus_multiproc_dir'] = tmpdir
```
bloodearnest: This is great - gives an ok fall back in dev or incase of misconfiguration, thanks!

docs/prometheus.rst
```diff
@@ -0,0 +1,39 @@
...
+Configuration
+-------------
+
+`prometheus_client` integration needs no extra configuration if running
+in single process mode.
+
+If the app is running in `multiprocess mode <https://github.com/prometheus/client_python#multiprocess-mode-gunicorn>`_
+(ie. with multiple workers), the `prometheus_multiproc_dir` envvar should be set
+to a writable directory that can be used for metrics. If this envvar is not found,
+a temporary directory will be created and used instead.
```
bloodearnest: Maybe clarify the fact that using the temp dir will break continuity between restarts?

Hmm, I wonder if we can use a non-random directory, say /tmp/talisker_multiproc_stats ? That way, the default fall back would only break continuity normally on reboot?

bloodearnest: Another thought - how does prometheus_client behave in a multithreaded environment? Is there locking involved?

talisker/gunicorn.py
maxiberta: Fixed.

talisker/gunicorn.py
maxiberta: `workers` is still not known at this point, that's why the envvar check was moved into `load_config()`; and the envvar is going to be set anyways... So I'm just checking if `prometheus_client` is installed.

docs/prometheus.rst
maxiberta: Non-random temp files and directories are subject to all kinds of timing attacks... We could probably use some smart deterministic algorithm, but not sure it's worth. Metrics will reset on reboots anyways, and it may not be as serious as we thought (see [this](https://www.robustperception.io/how-does-a-prometheus-counter-work/)).

If the envvar is not set, `prometheus_client` uses a simple `Lock` to update metrics in memory; so it should be thread-safe. If the envvar is set, it uses metrics protected by a mutex backed by a per-process mmaped file ([source code](https://github.com/prometheus/client_python/blob/master/prometheus_client/core.py)).

talisker/gunicorn.py
```diff
@@ -179,6 +194,24 @@ def init(self, parser, opts, args):
 
         return cfg
 
+    def load_config(self):
+        super(TaliskerApplication, self).load_config()
+
```
bloodearnest: at this point, self.cfg is the full config, including app defaults, config file, and cli args.

So, I think we should do the worker_exit hook here, rather than init. a) it allows us to inspect workers b) we can detect if the user has specified a custom worker_exit, and wrap it?

maxiberta: I'm not sure about b); I'd rather let users override the hook completely and do whatever they want with it.

bloodearnest: You might be right about b). I'm just concerned that that small comment in the docs about worker_exit might be missed. What's the worst case scenario if worker_exit doesn't clean up the dir?

PR #38, opened 2016-12-05
Better module caching

bloodearnest:
Improve the way we handle module level globals.

@module_cache allows a cached constructor function, which can be reset or bypassed for testing

module_dict() provides dict-based storage for module globals

context_local provides access to thread/greenlet local contexts that can also be cleared up when testing

Also, since this was a clean up, fix issue 26, and some pep8 fixes, and a fix rendering bug in the /_status/ index page, and standardize import conventions a bit more.


talisker/request_id.py
```diff
@@ -66,21 +66,27 @@ def decorator(id_func):
     def wrapper(func):
         @wraps(func)
         def decorator(*args, **kwargs):
+            current_id = get()
             id = id_func(*args, **kwargs)
+            if current_id == id:
```
emgee: I'm not really sure what this does. Perhaps a comment explaining why it short-circuits the func, i.e. does not set/restore the id, would be a good idea?

setup.cfg
```diff
@@ -12,6 +12,7 @@ test = pytest
 
 [flake8]
 exclude = .*,env,lib,dist,build
+ignore = E402
```
bloodearnest: I think it's a hangover from a previous commit. I noticed the flake 8 checks were broken anyway, so am removing the above and fixing that.

talisker/endpoints.py
```diff
@@ -47,7 +48,8 @@ def force_unicode(s):
     return s
 
 
-def load_networks():
+@module_cache
+def networks():
     networks = os.environ.get('TALISKER_NETWORKS', '').split(' ')
```
bloodearnest: Good call

talisker/request_id.py
bloodearnest: If the id is already set, it doesn't bother setting it.

talisker/util.py
```diff
@@ -22,6 +22,9 @@
 import pip
 from builtins import *  # noqa
 
+import functools
+
+import werkzeug.local
```
bloodearnest: werkzeug is a library, with some wsgi library stuff in it. werkzeug.local is a thread/greenlet context, independent of anything wsgi. Using talisker with celery already requires werkzeug, as its a core talisker dependency.

Changing this would mean moving the wsgi parts of talisker to an optional dependency, and vendoring werkzeug.local functionality. Or publishing a whole separate talisker_celery package or similar. Both of which have downsides, and are beyond this MP

talisker/util.py
```diff
@@ -35,3 +38,60 @@ def parse_url(url, proto='http'):
...
     return name in [x.project_name for x in pip.get_installed_distributions()]
+
+
+# a module level cache for global objects
+_global_cache = {}
+_global_dicts = []
+_context_locals = []
+
+
+def module_cache(func):
```
bloodearnest: GIL should protect here, I think?

talisker/util.py
emgee: The comment here was really just a counter to my, "This is good - celery shouldn't really depend on anything HTTP'y", comment above.

Although I always thought werkzeug was specifically an HTTP/WSGI utility library.

Not that it really matters - I agree it's beyond this MP.

talisker/util.py
emgee: Nope, sorry! Take `get` as an example:

1. Thread 1 executes `if id not in _global_cache`, sees that there's nothing in the cache, but is then interrupted.
2. Thread 2 jumps in, also sees nothing in the cache, allocates a new instance, o2, adds it to the cache, and returns o2 to the caller.
3. Thread 1 wakes up again and continues by allocating a new instance, o1, adding it to the cache, and returning o1 to the caller.

You now have one client with o1, one with o2, and a cache containing o1 (although it could just as easily be o2 in the cache, with o1 as the "floating" instance). Next time a client calls in, it will get the cached version but the service may be inconsistent by then.

Note: there are other points where an interrupt could occur, with similar effects.

But, like I said, maybe it doesn't actually matter for the current use case.

talisker/endpoints.py
```diff
@@ -56,29 +58,28 @@ def private(f):
     """Only allow approved source addresses."""
     @functools.wraps(f)
     def wrapper(self, request):
-        global NETWORKS, _loaded
```
bloodearnest: Deleted

talisker/util.py
bloodearnest: Hmm, I wonder if theres a different dict api I can use. I reluctant to lock the critical section, as that would be hit every single lookup

PR #39, opened 2016-12-06
Use celery signals to set request id automatically

bloodearnest:
Also, use task headers rather than thread local for celery enqueue metrics

### DIFF
--- a/talisker/celery.py
+++ b/talisker/celery.py
@@ -21,15 +21,14 @@ from __future__ import absolute_import
 
 from builtins import *  # noqa
 
-import functools
 import os
 import logging
+import time
 
 import talisker
 import talisker.logs
 import talisker.request_id
 import talisker.statsd
-from talisker.util import context_local
 
 
 __all__ = [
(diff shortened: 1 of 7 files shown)


talisker/celery.py
```diff
@@ -79,48 +52,35 @@ def signal(sender, **kwargs):
...
+    task.talisker_timer.start()
+    id = task.request.talisker_request_id
+    if id is not None:
+        talisker.request_id.set(id)
 
 
 def task_postrun(sender, task_id, task, **kwargs):
-    task.__talisker_timer.stop()
+    task.talisker_timer.stop()
+    talisker.request_context.clear()
```
emgee: Hmm, is there a talisker request_context to clear here? Where does the `postrun` signal execute - in the WSGI app or in the celery worker?

talisker/request_id.py
```diff
@@ -53,8 +53,12 @@ def get():
 
 def set(id):
     """Sets id in both general request context, and specific logging dict"""
-    request_context.request_id = id
-    set_logging_context(request_id=id)
+    if id is None:
+        del request_context.request_id
+        request_context.extra.pop('request_id'), None
```
emgee: I'm guessing ", None" should be inside the func's parentheses? Suggests there's a test missing.

talisker/request_id.py
```diff
@@ -53,8 +53,12 @@ def get():
 
 def set(id):
     """Sets id in both general request context, and specific logging dict"""
-    request_context.request_id = id
-    set_logging_context(request_id=id)
+    if id is None:
```
emgee: Setting to None and deleting feel like different operations but I'm fine if you're sure they're equivalent in the context of talisker.

talisker/celery.py
```diff
@@ -79,48 +52,35 @@ def signal(sender, **kwargs):
...
+        talisker.statsd.get_client().timing(name, ms)
 
 
 def task_prerun(sender, task_id, task, **kwargs):
     name = 'celery.{}.run'.format(sender.name)
-    task.__talisker_timer = talisker.statsd.get_client().timer(name)
-    task.__talisker_timer.start()
+    task.talisker_timer = talisker.statsd.get_client().timer(name)
+    task.talisker_timer.start()
+    id = task.request.talisker_request_id
```
bloodearnest: Yeah, this is celery dynamic programming madness. The task.request object proxies getattr to a dict, idk why </shrug>

talisker/celery.py
bloodearnest: It's an implicit context for the current request/job execution. It's either thread local storage, or greenlet local. Postrun runs in the celery worker, straight after the task completes.

Basically, the only way to provide request_id (and other stuff) to all logging calls is by sticking it in an implicit context. Sucks, but talisker's been that way from the start, as stdlib logging with it's globals is used everywhere.

talisker/request_id.py
bloodearnest: Well, if we set to None, we'll get "request_id=None" on every log line, so...

talisker/request_id.py
bloodearnest: Huh, yes, good catch

PR #42, opened 2016-12-21
Talisker networks logging

bloodearnest:
Log more info about 403 errors due to TALISKER_NETWORKS, and improved tests for X-Forwarded-For

Some other minor refactoring too.


talisker/endpoints.py
```diff
@@ -45,26 +46,46 @@ def force_unicode(s):
...
 
 @module_cache
-def networks():
-    networks = os.environ.get('TALISKER_NETWORKS', '').split()
-    return [ip_network(force_unicode(n)) for n in networks if n]
+def get_networks():
+    network_tokens = os.environ.get('TALISKER_NETWORKS', '').split()
+    network_list = []
+    for network in network_tokens:
+        network_list.append(ip_network(force_unicode(network)))
```
emgee: This could stay as a list comp I think?

emgee: Also, naming a var `xxx_list` grates a little ;). Just `networks` seems better.

talisker/endpoints.py
```diff
@@ -45,26 +46,46 @@ def force_unicode(s):
...
+    logger = logging.getLogger(__name__)
+    logger.info('configured TALISKER_NETWORKS',
+                extra={'networks': [str(n) for n in network_list]})
```
emgee: Should it log if there are no networks configured?

talisker/endpoints.py
```diff
@@ -45,26 +46,46 @@ def force_unicode(s):
...
 
 def private(f):
     """Only allow approved source addresses."""
+
     @functools.wraps(f)
     def wrapper(self, request):
         if not request.access_route:
-            # no client ip
-            return Response(status='403')
+            # this means something probably bugged in werkzeug, but lets fail
```
emgee: Picky, but I think it should be "let's" here.

talisker/endpoints.py
```diff
@@ -45,26 +46,46 @@ def force_unicode(s):
...
+    network_list = []
+    for network in network_tokens:
+        network_list.append(ip_network(force_unicode(network)))
+    logger = logging.getLogger(__name__)
+    logger.info('configured TALISKER_NETWORKS',
+                extra={'networks': [str(n) for n in network_list]})
+    return network_list
+
+
+PRIVATE_BODY_RESPONSE_TEMPLATE = """
```
emgee: Is there a test to prove the template is filled in correctly?

talisker/endpoints.py
bloodearnest: Good point, that was refactored and then re-refactored, I actually want to leave it alone now. Will do.

talisker/endpoints.py
bloodearnest: good catch, probably not

talisker/endpoints.py
bloodearnest: Nope, will add

PR #46, opened 2017-01-18
Add sentry support to celery

bloodearnest:
Celery job exceptions are now logged by sentry.

Add task id/name to logging context for celery jobs.

Refactor to use a sentry client subclass, as that was the best way
to automatically add things to the sentry context, for both wsgi and
celery.

Improve celery tests, including signal clean up.


talisker/sentry.py
```diff
@@ -91,27 +92,49 @@ def ensure_talisker_config(kwargs):
...
     if from_env:
         msg += ' from SENTRY_DSN environment'
-    logging.getLogger(__name__).info(msg, extra={'host': host})
+        extra['from_env'] = True
+    logging.getLogger(__name__).info(msg, extra=extra)
+
+
+def add_talisker_context(tags, extra):
+    if tags is None:
+            tags = {}
```
ricardokirkner: bad indentation

bloodearnest:
Have fixed the whitespace. Played around with a module level import, but it messes with the logging set up. I may revisit it at some point.

PR #47, opened 2017-01-20
Add django support, mainly sentry

bloodearnest:
Add support and docs for django. 
Includes a test django project in tests/django_app, which is mainly boilerplate

Also, moved flask support to use custom cls specification rather than the messy get_client(). It now works the same way django did. 

Also, enabled all signals in all celery contexts too, as I realise if a task is .apply()-ed, it runs (with signals) in the wsgi thread.


talisker/sentry.py
```diff
@@ -58,6 +59,11 @@ def register_client_update(update_func):
     return update_func
 
 
+def update_clients(client):
```
ricardokirkner: should this not better be called update_client (singular)?

talisker/django.py
```diff
@@ -0,0 +1,52 @@
...
+import logging
+from raven.contrib.django.client import DjangoClient
+
+import talisker.sentry
+
+
+# raven's django support does some very odd things.
+# There is a module global, that is a proxy to the output of
+# raven.contrib.django.models.get_client()
+# But the only way thing you can customise about the client is it's base class.
```
bloodearnest: yup, s/thing//

talisker/django.py
```diff
@@ -0,0 +1,52 @@
...
+
+class SentryClient(DjangoClient):
+    def __init__(self, *args, **kwargs):
+        # SQL hook sends raw SQL to the server. Not cool, bro.
+        # Force disable it for now, until we can add a processors to sanitize
+        kwargs['install_sql_hook'] = False
+        from_env = talisker.sentry.ensure_talisker_config(kwargs)
+        logging.getLogger(__name__).info(
+            'updating raven config from django app')
+        super().__init__(*args, **kwargs)
```
bloodearnest: It does work in python 2, via the future library, which provides a version for py2 that works.

talisker/sentry.py
bloodearnest: Well, you are updating all the references to the client, so I guess update_client_references?

talisker/django.py
bloodearnest: FYI, if the travis tests are passing, then they are passing under python 2.7, 3.4, 3.5

Issue #51, opened 2017-02-10
Enable logging for Django management commands


maxiberta:
Projects usually use management commands for cronjobs and other offline tasks. It'd be great if these scripts could take advantage of talisker's logging setup.

maxiberta:
Scripts can now be run with `python -m talisker` or `talisker.run`.

PR #53, opened 2017-02-14
Add run entrypoint for running scripts with talisker logging

bloodearnest:
Also works as -m talisker too.

Update: added docs, and fixed a niggling issue with docs that this MP triggered, move to a static version string in docs/conf.py, with bumpversion configured to update. Saves importing talisker in docs builds.

Fixes #53


talisker/__init__.py
```diff
@@ -33,3 +43,22 @@ def initialise():
...
     talisker.endpoints.get_networks()
     return devel
+
+
+def run():
+    """Initialise Talisker then exec python script."""
+    initialise()
+
+    if len(sys.argv) < 2:
+        sys.stderr.write('usage: python -m talisker <python script> ...')
```
maxiberta: Should exit here before crashing with an `IndexError`.

tests/test_main.py
```diff
@@ -0,0 +1,60 @@
...
+
+@pytest.fixture
+def script(tmpdir):
+    py_script = tmpdir.join('test.py')
+    py_script.write(SCRIPT)
+    return str(py_script)
+
+
+def test_run_entrypoint(script):
+    entrypoint = os.environ['VENV_BIN'] + '/' + 'talisker.run'
```
maxiberta: `os.path.join()` would probably be more correct here.

bloodearnest:
I don't think the import issue is anything to do talisker. It doesn't alter sys.path at all. Did it work when you ran the script by hand? If the script is in a subdir, you'd need to set pythonpath or similar anyway? Python only adds the current dir to sys.path, not the first the script is in, I think?

maxiberta:
Hm, it does work without `-m talisker`. Check this:
```
$ virtualenv /tmp/env
$ . /tmp/env/bin/activate
$ pip install -e .
$ echo "import b" > /tmp/a.py
$ echo "print('this is b')" > /tmp/b.py
$ python /tmp/a.py
this is b
$ python -m talisker /tmp/a.py
2017-02-14 17:18:14.879Z INFO talisker.sentry.TaliskerSentryClient "Raven is not configured (logging is disabled). Please see the documentation for more information."
2017-02-14 17:18:14.882Z INFO talisker.statsd "configuring statsd DummyClient"
Traceback (most recent call last):
  File "/usr/lib/python2.7/runpy.py", line 174, in _run_module_as_main
    "__main__", fname, loader, pkg_name)
  File "/usr/lib/python2.7/runpy.py", line 72, in _run_code
    exec code in run_globals
  File "/home/max/canonical/talisker/talisker/__main__.py", line 20, in <module>
    talisker.run()
  File "talisker/__init__.py", line 64, in run
    return exec_(code, globs, None)
  File "/tmp/env/local/lib/python2.7/site-packages/future/utils/__init__.py", line 634, in exec_
    exec("""exec code in globs, locs""")
  File "<string>", line 1, in <module>
  File "/tmp/a.py", line 1, in <module>
    import b
ImportError: No module named b
```
bewrwgbe

bloodearnest:
Ah, I bet -m changes cwd.

Does it work with talisker.run ?

maxiberta:
In the code example above, `/tmp` is missing from `sys.path` when invoked with `-m talisker`.

maxiberta:
Same error with `talisker.run`.

maxiberta:
This should fix issue #51 .

PR #57, opened 2017-03-15
support raven 6.0.0

bloodearnest:


bloodearnest:
I'd argue that version checking is even more fragile. I'd need to do different method signatures, and it may change again in 7.x. This way tests for behaviour, rather than versions.

I've tidied it up a bit, anyway, but I'm reluctant to explicitly check for version.

PR #70, opened 2017-04-10
Fix celery metrics with eager tasks

bloodearnest:
This PR fixes an issue with celery statsd metrics when the task is run synchronously/eagerly. It does so firstly by switching to raw timestamps rather than statsd.Timer objects, which fail when hit twice. It also changes the logic to cope with signal ordering in eager task execution, as in the presence of retries is very differerent from async tasks.

For a task that retries twice before failing, in async execution the signal order is

prerun
<task>
postrun
task_retry
prerun
<task retry 1>
postrun
task_retry
prerun
<task retry 2>
task_failure
postrun

However, if task is apply() synchronously/eagerly, the retry is recursive, and so it looks like this

prerun
<task>
prerun
<recursive task>
prerun
<recursive task>
task_failure
postrun
task_failure
postrun
task_failure
postrun

This pr adjusts the celery signal handlers that generate metrics to cope with both cases.

### DIFF
--- a/talisker/celery.py
+++ b/talisker/celery.py
@@ -46,10 +46,24 @@ def _counter(name):
     return signal
 
 
+def _protected_counter(name):
+    """Count metrics, but ensure only once.
+
+    This is needed when tasks are eagerly invoked and have reties, or else
+    metrics will be duplicated."""
+    attr = '_talisker_sent_' + name
+
+    def protected_signal(sender, **kwargs):
+        if not hasattr(sender, attr):
+            stat_name = 'celery.{}.{}'.format(sender.name, name)
+            talisker.statsd.get_client().incr(stat_name)
+    return protected_signal
+
+
 task_retry = _counter('retry')
-task_success = _counter('success')
-task_failure = _counter('failure')
-task_revoked = _counter('revoked')
+task_success = _protected_counter('success')
+task_failure = _protected_counter('failure')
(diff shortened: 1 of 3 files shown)


talisker/celery.py
```diff
@@ -88,18 +103,29 @@ def after_task_publish(sender, body, **kwargs):
...
 def task_prerun(sender, task_id, task, **kwargs):
-    name = 'celery.{}.run'.format(sender.name)
-    task.talisker_timer = talisker.statsd.get_client().timer(name)
-    task.talisker_timer.start()
     rid = get_header(task.request, REQUEST_ID)
     if rid is not None:
         talisker.request_id.push(rid)
+    if hasattr(task, 'talisker_timestamp') and task.request.is_eager:
+        # eager task, but retry has happened immeadiately, so send metrics
+        send_run_metric(sender.name, task.talisker_timestamp)
```
maxiberta: Retried eager tasks send timings twice, on prerun and on postrun. Is that ok?

bloodearnest: yeah, that's the goal. As when eager, all preruns are fired before any postruns are called.

So the result of this PR is that in that case, the 2nd+ preruns send metrics for the retry, and the first postrun sends it for the final retry. The other postruns with then fire do nothing.

It's awkward, but it's what celery does, so..

PR #71, opened 2017-04-10
Fix statsd cli args and metric format

bloodearnest:
Also, avoid blowing up if on old setuptools

### DIFF
--- a/talisker/gunicorn.py
+++ b/talisker/gunicorn.py
@@ -27,7 +27,7 @@ import tempfile
 import sys
 from collections import OrderedDict
 
-from gunicorn.instrument import statsd as gstatsd
+from gunicorn.glogging import Logger
 from gunicorn.app.wsgiapp import WSGIApplication
 
 from . import logs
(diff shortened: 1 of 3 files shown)


talisker/gunicorn.py
```diff
@@ -58,16 +58,8 @@ def prometheus_multiprocess_worker_exit(server, worker):
...
 
 
-class GunicornLogger(gstatsd.Statsd):
-    """Custom gunicorn logger to use structured logging.
-
-    Based on the statsd gunicorn logger, and also increases timestamp
-    resolution to include msec in access and error logs.
-    """
-    def __init__(self, cfg):
-        super(GunicornLogger, self).__init__(cfg)
```
maxiberta: This fixes the statsd cli crashes but, can we just drop it with no other consequences?

bloodearnest: Yes, we the only part we were still using was the log metrics, which I moved the our class.

talisker/gunicorn.py
```diff
@@ -58,16 +58,8 @@ def prometheus_multiprocess_worker_exit(server, worker):
...
-        if self.sock is not None:
-            self.sock.close()
```
bloodearnest: Yeah, we didn't use it. The base class opened a socket, but we didn't use it, so I just closed it to make sure.

This change just drops the use of that base class altogether, as we have diverged/customised significantly

PR #80, opened 2017-04-21
Refactor devel and colored output

### LINKED ISSUE #77: Talisker 0.9.3 inserts \x1b into log files
As it says on the tin. We're running:

talisker 0.9.3
python 3.4.3

deployed on trusty.

We think this was introduced when talisker added automatic colouration when stderr is a tty. Perhaps for some reason talisker thinks an upstart-owned stderr is a tty, and so applies coloration to it's log output?

bloodearnest:
This PR:

a) reverts to requiring DEVEL env var to enable colors, as this is safer, and upstart uses ptys, so broke our attempt at detection

b) color can always be disabled by setting TALISKER_NO_COLOR envar

c) allows explicitly disabling devel with DEVEL=False

It also adds a couple of missing tests.

Fixes #77

### DIFF
--- a/talisker/__init__.py
+++ b/talisker/__init__.py
@@ -54,10 +54,12 @@ ACTIVE = set(['true', '1', 'yes'])
 def get_config(env=os.environ):
     """Load talisker config from environment"""
     devel = env.get('DEVEL', '').lower() in ACTIVE
-    if 'TALISKER_COLOR' in env:
-        color = env.get('TALISKER_COLOR', '').lower() in ACTIVE
-    else:
-        color = devel and sys.stderr.isatty()
+    color = False
+    if devel:
+        if 'TALISKER_COLOR' in env:
+            color = env.get('TALISKER_COLOR', '').lower() in ACTIVE
+        else:
+            color = sys.stderr.isatty()
 
     return {
         'devel': devel,
--- a/talisker/__init__.py
+++ b/talisker/__init__.py
@@ -54,12 +54,15 @@ ACTIVE = set(['true', '1', 'yes'])
 def get_config(env=os.environ):
     """Load talisker config from environment"""
     devel = env.get('DEVEL', '').lower() in ACTIVE
+    if 'TALISKER_COLOR' in env:
+        color = env.get('TALISKER_COLOR', '').lower() in ACTIVE
+    else:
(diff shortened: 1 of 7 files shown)


talisker/__init__.py
```diff
@@ -48,11 +48,18 @@ def initialise(env=os.environ):
...
+
 def get_config(env=os.environ):
     """Load talisker config from environment"""
+    devel = env.get('DEVEL', '').lower() in ACTIVE
     return {
+        'devel': devel,
         'debuglog': env.get('DEBUGLOG'),
-        'devel': sys.stderr.isatty() or 'DEVEL' in env,
+        'color': (
+            devel and sys.stderr.isatty() and 'TALISKER_NO_COLOR' not in env
```
emgee: Would it be better to have a `TALISKER_COLOR` env var that can be either "yes" or "no" and, if present, use that in preference to any tty detection? That would mean you could also override a false -ve from `isatty()`.

talisker/gunicorn.py
```diff
@@ -219,7 +219,7 @@ def load_config(self):
                 extra={'errorlog': self.cfg.errorlog})
             self.cfg.set('errorlog', '-')
 
-        if self.cfg.loglevel.lower() == 'debug':
+        if self.cfg.loglevel.lower() == 'debug' and self._devel:
```
emgee: I probably don't have full context here ... but is there a reason to block "debug" level logs just because it's not a devel system?

tests/test_init.py
```diff
@@ -24,17 +24,20 @@
 import talisker
 
 
+def assert_config(env, **kwargs):
```
emgee: Totally optional, but I'd probably s/kwargs/expected to give the param name proper meaning.

talisker/gunicorn.py
bloodearnest: Yes, we don't ever want to send DEBUG logs to stderr in prod, as they are a) sensitive and b) voluminous, so we force devel mode

tests/test_init.py
bloodearnest: Done

talisker/__init__.py
bloodearnest: Done

PR #81, opened 2017-04-24
fix some grok filter issues

bloodearnest:
Allow _ in logger name, and handle any foo= values by setting them to empty string, as the kv filter can't parse them properly

### DIFF
--- a/docs/logging.rst
+++ b/docs/logging.rst
@@ -209,11 +209,11 @@ line, following the `logfmt <https://brandur.org/logfmt>`_ idea. e.g.::
 
 .. sidebar:: Defining logfmt
 
-    logfmt is very loosely specified, and our target parser has some limitations,
-    so we define it as:
+    logfmt is very loosely specified, and our target parser (logstash) has some
+    limitations, so we define it as:
 
     * keys: any string, except:
-        - `` ``, ``.``, and ``=`` are replaced by ``_``
+        - `` ``, ``.``, and ``=`` are replaced by ``_``, due to logstash limitations.
         - ``"`` is replaced by ``""``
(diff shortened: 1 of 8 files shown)


docs/logging.rst
```diff
@@ -209,24 +209,26 @@ line, following the `logfmt <https://brandur.org/logfmt>`_ idea. e.g.::
...
 
-    Both keys and values can be of arbitrary length, and either utf8 encoded
-    bytes, or unicode. Talisker will always encode the output in utf8.
+    Both keys and values are encoded in utf8, and truncated after 2kb, to avoid
+    accidental hugh log messages.
 
-    The reason for stripping " characters is to do with the limitations of
-    logstash's kv filter, which cannot currently cope with them, even when
-    escaped. See `issue 2
+    The
```
wgrant: Not meant to be here?

talisker/logs.py
```diff
@@ -318,32 +325,41 @@ def remove_quotes(self, s):
...
-        if isinstance(k, bytes):
-            k = k.decode('utf8')
+        # we need unicode strings so as to be able to do replace
+        if not v:
+            return None
+
         if not isinstance(v, str):
             if isinstance(v, bytes):
                 v = v.decode('utf8')
             elif not isinstance(v, str):
```
wgrant: Isn't this just "else", given it's already inside an if block with the same predicate?

docs/logging.rst
bloodearnest: yep, gone

talisker/logs.py
bloodearnest: Indeed it is, gone

Issue #84, opened 2017-04-26
Support PyPy


bloodearnest:
We used to support pypy, but the introduction of psycopg2 support causes issues, so we have removed it.

We should be able to leverage psycopg2cffi as a drop in, but managing the intstall_requires needs some care. Also, the pytest-postgresql test plugin we use doesn't support using psycopg2cffi, so we'd need to fork/upstream a change that allows it to depend on/use pyscopg2cffi on pypy.

jugmac00:
I do not think supporting pypy is any more a valid requirement. If so, please re-open the issue.

PR #85, opened 2017-04-26
add support for postgresql via psycopg2

bloodearnest:
This PR adds optional integration with psycopg2, in the form of a custom connection class

If used, it provides

a) add sql queries to sentry reports
b) slow queries application logs

It also extends the test django and flask apps to use a postgresql db, and adds a setuptools extra for installing postgresql dependencies. We also drop support for pypy, as that requires a fair bit more work, see #84.


talisker/postgresql.py
```diff
@@ -0,0 +1,142 @@
...
+        query = prettify_sql(query)
+        query = FILTERED if query is None else query
+        duration = '{:d}ms'.format(int(math.ceil(t)))
+        connection = get_safe_connection_string(self)
+        return query, duration, connection
+
+    def _record(self, msg, query, duration):
+        query_data = None
+
+        if self.query_mintime >= 0 and duration > self.query_mintime:
```
ricardokirkner: given this usage, should the variable not be best called query_maxtime?

talisker/postgresql.py
```diff
@@ -0,0 +1,142 @@
...
+
+    @property
+    def query_mintime(self):
+        if self._mintime is None:
+            self._mintime = talisker.get_config()['slowquery_time']
+        return self._mintime
+
+    def cursor(self, *args, **kwargs):
+        kwargs.setdefault('cursor_factory', TaliskerCursor)
+        return super().cursor(*args, **kwargs)
```
ricardokirkner: can you use super() in Python 2.7?

tests/django_app/manage.py
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+#!../../env/bin/python
```
ricardokirkner: possibly not for this PR but wouldn't it be better to move the django_app outside of the tests folder?

talisker/postgresql.py
bloodearnest: Hmm, perhaps, I don't mind either way, so I will rename

talisker/postgresql.py
bloodearnest: You can with python-future.org :) It's a bit lazy of me, as being explicit is supported in both, but I am so. very. tired. of explicit super in py2, so I just go all out and rely future's py2 workaround.

tests/django_app/manage.py
bloodearnest: perhaps, but where would it live? It (and related flask/wsgi/celery apps) would need to be in another top level dir, not sure I see value in that.

FWIW, I explicitly exclude it from the test search path in pytest (see setup.cfg/norecursedirs)

talisker/postgresql.py
ricardokirkner: nice... lazyness is good

PR #86, opened 2017-04-26
truncate long log messages

bloodearnest:
Truncates long logmsg and logfmt parts of log lines

Last line of defence, as Talisker now truncates these at generation time, but just in case.

Also refactored some of the logstash test setup a bit.


talisker/logstash/talisker.filter
```diff
@@ -5,9 +5,17 @@ filter {
         negate => true
         what => "previous"
     }
+    ruby {
+      code => "
+        if event['message'].length > 20000
+            event.tag 'message truncated'
+            event['message'] = event['message'][0..20000] + '...'
```
wgrant: You have an off-by-one error here. 0..20000 is 20001 elements, so a string with 20001 elements will be marked as truncated but not actually be.

PR #90, opened 2017-04-27
Clear sentry context on request start

bloodearnest:
Otherwise it bleeds into the request afterwards, which is super confusing

### DIFF
--- a/talisker/sentry.py
+++ b/talisker/sentry.py
@@ -53,10 +53,6 @@ default_processors = set([
 
 sentry_globals = module_dict()
 
-# access logs are generated after a request has finished and the breadcrumb
-# context cleared. This means that they appear in the next requests
-# breadcrumbs, unless we ignore it
-raven.breadcrumbs.ignore_logger('gunicorn.access')
 # sql queries are recorded as breadcrumbs anyway, so don't record them as log
 # queries twice if they happen to be slow
 raven.breadcrumbs.ignore_logger('talisker.slowqueries')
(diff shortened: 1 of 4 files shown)


talisker/sentry.py
```diff
@@ -53,6 +53,10 @@
 
 sentry_globals = module_dict()
 
+# access logs a generated after a request has finished and the breadcrumb
```
Roadmaster: access logs "are" ?

bloodearnest: fixed, thanks!

bloodearnest:
hmm this is not sufficient. Need a better approach to insuring no misleading logs from other requests are caught up

tests/test_sentry.py
```diff
@@ -148,13 +148,33 @@ def test_log_client(monkeypatch, log):
...
+
+    def app(environ, sr):
+        logger.info('app log')
+        return []
+
+    mw = talisker.sentry.get_middleware(app)
+    context = mw.client.context
+    context.clear()  # clear startup logs
+    assert len(context.breadcrumbs.buffer) == 0
+    mw(environ, lambda: None)
```
fgallina: As the middleware itself should clear the contents I don't think l.167 is really required, moreover IIUC two or more calls of this kind (`mw(environ, lambda: None)`) should result in the assertions below still as still passing as the context shall be cleared every-time, maybe it'd be a good idea to just test this that way.

bloodearnest: Fair point, done

Issue #93, opened 2017-05-18
Multiline messages overflow into traceback (grok)


squidsoup:
Multiline log messages appear to overflow into 'traceback' resulting in `status`, `ip` etc. not being logged. This is quite misleading as we often filter out 200 `status` to look for issues.

See: https://logging.ols.canonical.com/app/kibana#/doc/logstash-*/logstash-2017.05.18/talisker?id=AVwdUQf2ZwZISI6xIgcV&_g=(refreshInterval:(display:Off,pause:!f,value:0),time:(from:now-15m,mode:quick,to:now))

Hopefully addressed by https://github.com/canonical-ols/talisker/issues/87 but worth confirming.

bloodearnest:
Hmm, was there a \n in the query string, perhaps? Talisker should probably strip any '\n' chars from the logmsg before sending

squidsoup:
Not certain, but there are a few similar events with the overflow in Kibana
for other queries. Let me know if I can help with surfacing other
misbehaving ones.

PR #94, opened 2017-05-22
Access log cleanups

bloodearnest:
This strips newlines from the logmsg, as it confuses grok. Also, it doesn't include the querystring in the logmsg for gunicorn's access log, as it can be overly long and is already included in the qs field

### DIFF
--- a/talisker/gunicorn.py
+++ b/talisker/gunicorn.py
@@ -61,17 +61,11 @@ class GunicornLogger(Logger):
     """Custom gunicorn logger to use structured logging."""
 
     def get_extra(self, resp, req, environ, request_time):
-        status = resp.status
-        if isinstance(status, (str, bytes)):
-            status = status[:3]
-        else:
-            status = str(status)
-
         extra = OrderedDict()
         extra['method'] = environ.get('REQUEST_METHOD')
         extra['path'] = environ.get('PATH_INFO')
         extra['qs'] = environ.get('QUERY_STRING')
-        extra['status'] = status
+        extra['status'] = str(resp.status_code)
         extra['ip'] = environ.get('REMOTE_ADDR', None)
         extra['proto'] = environ.get('SERVER_PROTOCOL')
         extra['length'] = getattr(resp, 'sent', None)
@@ -124,12 +118,10 @@ class GunicornLogger(Logger):
             request_time.seconds * 1000 +
             float(request_time.microseconds) / 10 ** 3
         )
-        status = resp.status
-        if isinstance(status, (str, bytes)):
-            status = int(status.split(None, 1)[0])
         self.histogram("gunicorn.request.duration", duration_in_ms)
         self.increment("gunicorn.requests", 1)
-        self.increment("gunicorn.request.status.{}".format(status), 1)
+        self.increment("gunicorn.request.status.{}".format(
+            resp.status_code), 1)
 
     def setup(self, cfg):
         super(GunicornLogger, self).setup(cfg)
@@ -213,7 +205,7 @@ class TaliskerApplication(WSGIApplication):
 
         # override and warn
         if self.cfg.errorlog != '-':
-            logger.warn(
+            logger.warning(
                 'ignoring gunicorn errorlog config, talisker logs to stderr',
                 extra={'errorlog': self.cfg.errorlog})
             self.cfg.set('errorlog', '-')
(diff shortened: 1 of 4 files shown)


bloodearnest:
This should probably fix #93

talisker/gunicorn.py
```diff
@@ -90,6 +89,9 @@ def get_extra(self, resp, req, environ, request_time):
         if request_id:
             extra['request_id'] = request_id
 
+        msg = "{} {}".format(extra['method'], extra['path'])
```
sparkiegeek: minor nit, I'd spell this as:

`msg = "{0.method} {0.path}".format(**extra)`

talisker/logs.py
```diff
@@ -282,7 +282,7 @@ def __init__(self, fmt=None, datefmt=None):
 
     def format(self, record):
         """Format message, with escaped quotes and structured tags."""
```
sparkiegeek: docstring is out of date now

tests/test_gunicorn.py
```diff
@@ -121,14 +122,29 @@ def test_gunicorn_logger_access(environ, log, statsd_metrics):
     cfg.set('accesslog', '-')
     logger = gunicorn.GunicornLogger(cfg)
 
+    log[:] = []
```
sparkiegeek: `log.clear()` seems clearer (pun intended) :wink:

tests/test_gunicorn.py
```diff
@@ -121,14 +122,29 @@ def test_gunicorn_logger_access(environ, log, statsd_metrics):
...
 
 
+def test_gunicorn_logger_access_qs(environ, log):
+    response, environ, delta, expected = access_extra_args(
+        environ, '/url?foo=bar')
+    cfg = Config()
+    cfg.set('accesslog', '-')
+    logger = gunicorn.GunicornLogger(cfg)
+
+    log[:] = []
```
sparkiegeek: `clear()`?

tests/test_gunicorn.py
```diff
@@ -139,6 +155,7 @@ def test_gunicorn_logger_access_with_request_id(environ, log):
     cfg.set('accesslog', '-')
     logger = gunicorn.GunicornLogger(cfg)
 
+    log[:] = []
```
sparkiegeek: `clear()`?

talisker/logs.py
```diff
@@ -323,8 +323,8 @@ def format(self, record):
                                                'replace')
         return s
 
-    def escape_quotes(self, s):
-        return s.replace('"', '\\"')
+    def clean_message(self, s):
+        return s.replace('"', '\\"').replace('\n', '')
```
sparkiegeek: Solving a similar problem with syslog in the past, we replaced \n with `#012`

I'm a bit concerned that just throwing it away will cause misleading logs - if we had something blow up on newlines, how would we be able to identify that request from these logs?

tests/test_gunicorn.py
bloodearnest: Yeah, that's what I wrote originally, but no py2 support... :(

sparkiegeek: Ahh, fair enough :disappointed:

talisker/gunicorn.py
```diff
@@ -90,6 +89,9 @@ def get_extra(self, resp, req, environ, request_time):
...
+        if extra['qs']:
+            msg += '?'
```
emgee: The '?' is just an indicator there's also a query string to look at, right? Do you think it's worth including? If so, maybe it's worth bundling in the `format` call above to avoid the extra string append.

talisker/logs.py
emgee: +1 to concern about throwing the existence of the '\n' away.

tests/test_gunicorn.py
emgee: Could also `del log[:]` but I'm not sure that's any clearer (pub also intended).

talisker/gunicorn.py
bloodearnest: Done (with corrections)

talisker/gunicorn.py
```diff
@@ -61,12 +61,11 @@ class GunicornLogger(Logger):
...
 
     def get_extra(self, resp, req, environ, request_time):
-
-        msg = "%s %s" % (environ['REQUEST_METHOD'], environ['RAW_URI'])
-
         status = resp.status
         if isinstance(status, (str, bytes)):
             status = status[:3]
+        else:
+            status = str(status)
```
bloodearnest: Yeah, I just realised the it was relying on upstream str() conversion to work, so I slipped it in.
Bizarrely, in some circumstances, status is an int, but I've not been able to get to the bottom of why.

talisker/gunicorn.py
bloodearnest: I do think it's worth including, or else I wouldn't have included it ;P

It's what other tools do (e.g. squid) when logging similar urls. The purpose is to give a quick indicator to a reader that this url had a query string, signalling that they may need to look deeper. Part of the issue was that the querystrings can be very long, and they are already included in the structured data.

PR #99, opened 2017-05-31
Workarounds for py36/async/requests bug

### LINKED ISSUE #98: Requests error during ssl with eventlet
Hey, thanks for your help with my previous noob question with PyCharm and debugging. Your link totally worked. Funny i didn't know about that before because i have never been able to dev in the same env as prod. Which is why i have been so excited about `talisker`. Hopefully will finally catch some prod env bugs before deployment.

I pretty much got my entire setup refactored to using `talisker` which was awesome and started testing some things. There seems to be an error when making any external https requests. This time i have a small test case for you, i just hope it isn't `macOS` related. The error isn't surfaced so i don't have that.

Command is `talisker -k eventlet taltest:app`.
Changing to `gunicorn -k eventlet taltest:app` does not error so i hoped you would know how to surface the error and then it would be an easy fix.

```python
#!/usr/bin/env python
# -*- coding: utf-8 -*-

"""Talisker repro with ssl and requests on py 3.6

eventlet==0.21.0
Flask==0.12.2
talisker==0.9.5
"""

from flask import Flask
app = Flask(__name__)

import requests
print(requests.get("https://google.com"))
```

One weird thing is i changed `talisker.gunicorn.run` to below just to play with things. And the error didn't go away. Which stumped me since i copied the lines from the `gunicorn.wsgiapp.run` method.

```
def run():
  # config = talisker.initialise()
  # talisker.celery.enable_signals()
  # return TaliskerApplication(
  #     "%(prog)s [OPTIONS] [APP_MODULE]", config['devel'], config['debuglog'])
  from gunicorn.app.wsgiapp import WSGIApplication
  return WSGIApplication("%(prog)s [OPTIONS] [APP_MODULE]").run()
```

bloodearnest:
Work around for https://github.com/requests/requests/issues/3752

Adds two new special purpose entrypoints to do early monkeypatching to avoid breaking https requests with py36 and eventlet/gevent

Also refactor entrypoint functions into \_\_init__.py, consolidating into one place and making import management easier.

Fixes #98


talisker/logs.py
```diff
@@ -50,6 +50,10 @@ def set_global_extra(extra):
 
 def reset_logging():
     """Reset logging config"""
+    # avoid unclosed file resource warning
+    for handler in logging.getLogger().handlers:
+        if getattr(handler, '_debug_handler', False):
+            handler.stream.close()
```
sparkiegeek: hmm, this seems unrelated to the change? can we pull it out to a separate PR?

bloodearnest:
Have rebased and addressed comments.

PR #100, opened 2017-06-05
Add talisker.testing module, with testing helpers

bloodearnest:
Includes:

 * helper for parsing and asserting against Talisker logging
 * helper for running gunicorn for integration tests

Also, I had to fix the status_code bug to test the helper, will follow
up with PR with integration test coverage.

### DIFF
--- a/talisker/testing.py
+++ b/talisker/testing.py
@@ -226,7 +226,8 @@ class GunicornProcess(ServerProcess):
         super().__enter__()
         self.output.append(self.ps.stdout.readline())
         while self.WORKER not in self.output[-1]:
-            self.check()
+            with self._handle_enter_exception():
+                self.check()
             m = self.ADDRESS.search(self.output[-1])
             if m:
                 self.port = m.groups()[0]
--- a/talisker/testing.py
+++ b/talisker/testing.py
@@ -1,0 +1,260 @@
+# Copyright (C) 2016- Canonical Ltd
+#
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 2 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program; if not, write to the Free Software
+# Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA
+
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+from __future__ import print_function
+from __future__ import division
+from __future__ import absolute_import
+
+from builtins import *  # noqa
+import sys
+import subprocess
+import re
+import shlex
+import logging
+from contextlib import contextmanager
+import weakref
+
+import requests
+
+import talisker.logs
+
+
+class TestLogger(talisker.logs.StructuredLogger):
+
+    def __init__(self, name):
+        super().__init__(name)
+        self.records = []
+        self.lines = []
+
+
+class TestHandler(logging.Handler):
+    """Testing handler that records its logs in memory."""
+    def __init__(self, logger, level=logging.NOTSET):
+        assert isinstance(logger, TestLogger)
+        super().__init__(level)
+        self.logger = weakref.proxy(logger)
+
+    def emit(self, record):
+        self.logger.records.append(record)
+        self.logger.lines.extend(self.format(record).split('\n'))
+
+
+def test_logger():
+    logger = TestLogger('test')
+    handler = TestHandler(logger)
+    handler.setFormatter(talisker.logs.StructuredFormatter())
+    handler.setLevel(logging.NOTSET)
+    logger.addHandler(handler)
+    return logger
+
+
+class LogOutput:
+    """A container for log messages output by a Talisker program.
+
+    Parses log lines, and makes them available for inspection in tests.
+    """
+
+    TIMESTAMP = re.compile(r'^\d\d\d\d-\d\d-\d\d \d\d:\d\d:\d\d.\d\d\dZ')
+
+    def __init__(self, lines):
+        self.logs = list(self.read(lines))
+
+    def read(self, lines):
+        current = []
+        for line in lines:
+            if self.TIMESTAMP.match(line):
+                if current:
+                    yield self.parse(current)
+                current = []
+            current.append(line)
+        if current:
+            yield self.parse(current)
+
+    def parse(self, logs):
+        """Stupid simple logfmt parser"""
+        log = logs[0]
+        trailer = logs[1:]
+        parsed = shlex.split(log)
+        date, time, level, name, msg = parsed[:5]
+        try:
+            extra = dict((v.split('=')) for v in parsed[5:])
+        except:
+            assert 0, "failed to parse logfmt: " + log
+        return {
+            'ts': date + " " + time,
+            'level': level,
+            'logger': name,
+            'logmsg': msg,
+            'extra': extra,
+            'trailer': trailer,
+        }
+
+    def _compare_strings(self, needle, haystack):
+        return (
+            all(k in haystack for k in needle) and
+            all(needle[k] in haystack[k] for k in needle)
+        )
+
+    def exists(self, extra=None, trailer=None, **kwargs):
+        for log in self.logs:
+            strings_match = self._compare_strings(kwargs, log)
+            if extra is not None:
+                extra_match = self._compare_strings(extra, log['extra'])
+            else:
+                extra_match = True
+            if trailer is not None:
+                trailer_match = True
+                for line in trailer:
+                    if not any(line in tline for tline in log['trailer']):
+                        trailer_match = False
+                        break
+            else:
+                trailer_match = True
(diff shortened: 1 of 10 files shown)


bloodearnest:
This will be followed up by other PRs:

1) use the new helpers in the existing code
2) write some new integration tests using the new helpers
3) add helpers to talisker.testing for statsd/logging capture
4) update documentation

bloodearnest:
Also, a limitation of this approach is that pipes have a buffer size, so a user could block if they write large amounts of stdout/err in their subprocess w/o reading it.

Athough a bit more complicated, it might be better to output to file, rather than use a pipe.

talisker/testing.py
```diff
@@ -0,0 +1,265 @@
...
+import logging
+from contextlib import contextmanager
+import weakref
+
+import requests
+
+import talisker.logs
+
+
+class TestLogger(talisker.logs.StructuredLogger):
```
atomatt: Just wondering if some of the classes in this module are internal implementation and would be better with a underscore prefix?

tests/test_testing.py
```diff
@@ -0,0 +1,87 @@
...
+        'logmsg': 'msg 2',
+        'extra': {'foo': 'bar', 'baz': '1'}
+    } not in log
+    assert {'logmsg': 'msg 3'} in log
+    assert {'logmsg': 'msg 3', 'trailer': ['line1']} in log
+    assert {'logmsg': 'msg 3', 'trailer': ['line2']} in log
+
+
+def test_serverprocess_success():
+    server = testing.ServerProcess(['true'])
```
atomatt: Nice :)

tests/test_testing.py
```diff
@@ -0,0 +1,87 @@
...
+
+def test_serverprocess_exit():
+    server = testing.ServerProcess(['false'])
+    with server:
+        with pytest.raises(testing.ServerProcessError):
+            while 1:
+                server.check()
+
+
+def test_guicornprocess_success():
```
atomatt: guicorn now? :D

talisker/testing.py
bloodearnest: No, it's meant to be used externally too, if needed. You might be doing non server stuff that needs to parse log output.

talisker/testing.py
```diff
@@ -0,0 +1,265 @@
...
+        self.records = []
+        self.lines = []
+
+
+class TestHandler(logging.Handler):
+    """Testing handler that records its logs in memory."""
+    def __init__(self, logger, level=logging.NOTSET):
+        assert isinstance(logger, TestLogger)
+        super().__init__(level)
+        self.logger = weakref.proxy(logger)
```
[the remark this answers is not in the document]
bloodearnest: I'm cheating by giving the handler a reference to it's owning Logger, which would be a reference cycle. The reason for that is so the handler can store the log messages on the logger.

There may be a cleaner way to do this. Instead of the reference, the handler could just store the log messages iself, and TestLogger could have a property that gets them from it's handlers.

talisker/testing.py
```diff
@@ -0,0 +1,265 @@
...
+            yield self.parse(current)
+
+    def parse(self, logs):
+        """Stupid simple logfmt parser"""
+        log = logs[0]
+        trailer = logs[1:]
+        parsed = shlex.split(log)
+        date, time, level, name, msg = parsed[:5]
+        try:
+            extra = dict((v.split('=')) for v in parsed[5:])
```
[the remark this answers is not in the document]
bloodearnest: hmm, probably yes

talisker/testing.py
```diff
@@ -0,0 +1,265 @@
...
+        except:
```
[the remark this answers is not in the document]
bloodearnest: ok

talisker/testing.py
```diff
@@ -0,0 +1,265 @@
...
+            'logmsg': msg,
+            'extra': extra,
+            'trailer': trailer,
+        }
+
+    def _compare_strings(self, needle, haystack):
+        return (
+            all(k in haystack for k in needle) and
+            all(needle[k] in haystack[k] for k in needle)
+        )
```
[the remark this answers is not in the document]
bloodearnest: yeah, that looks better

talisker/testing.py
```diff
@@ -0,0 +1,265 @@
...
+
+    def exists(self, extra=None, trailer=None, **kwargs):
```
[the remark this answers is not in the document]
bloodearnest: Totally is leaking implementation, will fix

talisker/testing.py
```diff
@@ -0,0 +1,265 @@
...
+
+
+class ServerProcessError(Exception):
+    pass
+
+
+class ServerProcess(object):
+    """Context mananger to run a server subprocess """
+
+    _log = None
```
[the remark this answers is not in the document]
bloodearnest: ok

talisker/testing.py
```diff
@@ -0,0 +1,265 @@
...
+        # might be a way to include it in the exception somehow
+        if type is ServerProcessError:
+            sys.stderr.write('Server process died:\n')
+            sys.stderr.write(''.join(self.output))
+
+
+class GunicornProcess(ServerProcess):
+    """Context mananger to run Talisker's gunicorn server.
+
+    It captures all output, and waits till it sees gunicorn which port
```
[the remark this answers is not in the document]
bloodearnest: ok

talisker/testing.py
```diff
@@ -0,0 +1,265 @@
...
+            all(needle[k] in haystack[k] for k in needle)
+        )
+
+    def exists(self, extra=None, trailer=None, **kwargs):
+        for log in self.logs:
+            strings_match = self._compare_strings(kwargs, log)
+            if extra is not None:
+                extra_match = self._compare_strings(extra, log['extra'])
+            else:
+                extra_match = True
```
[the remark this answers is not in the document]
bloodearnest: kk

tests/test_testing.py
bloodearnest: gah

talisker/testing.py
```diff
@@ -0,0 +1,265 @@
...
+        if rc is not None and rc > 0:
+            raise ServerProcessError('subprocess errored')
+
+    @contextmanager
+    def _handle_enter_exception(self):
+        """Helper to manually handle an exception in the __enter__ method"""
+        try:
+            yield
+        except:
+            suppressed = self.__exit__(*sys.exc_info())
```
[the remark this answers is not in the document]
bloodearnest: So, this is a bit tricky, and probably deserves a comment.

If an error is raised during a context manager's __enter__ method, python will not invoke it's __exit__ method, as the context manager hasn't started yet - it just bubbles up like a regular exception.

But we want to do some clean up in this case (kill the process, read the logs), so we need the clean up behaviour in __enter__ too.

To make it clearer, I've renamed some stuff and added some more comments. But basically we are manually adding the clean up code to the risky parts of __enter__

talisker/testing.py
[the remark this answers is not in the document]
bloodearnest: hmm, this doesn't work when I replace it. Not sure why.

talisker/testing.py
```diff
@@ -0,0 +1,273 @@
...
+        trailer = match.pop('trailer', None)
+        for log in self.logs:
+            strings_match = self._compare_strings(match, log)
+            extra_match = trailer_match = True
+
+            if extra is not None:
+                extra_match = self._compare_strings(extra, log['extra'])
+
+            if trailer is not None:
+                trailer_match = True
```
atomatt: I think this line is redundant.

talisker/testing.py
```diff
@@ -0,0 +1,273 @@
...
+            sys.stderr.write('Server process died:\n')
+            sys.stderr.write(''.join(self.output))
+
+    @contextmanager
+    def _handle_enter_exception(self):
+        """Helper to manually handle an exception in the __enter__ method"""
+        try:
+            yield
+        except:
+            suppressed = self.handle_exception(*sys.exc_info())
```
atomatt: `handle_exception` always returns None, so the conditional on the next line is not meaningful. I think an unexpected server process exit should be supressed?

talisker/testing.py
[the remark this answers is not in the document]
atomatt: As discussed, not a big issue because the dicts should be quite small and test code is generally not performance-critical anyway.

talisker/testing.py
```diff
@@ -0,0 +1,273 @@
...
+    def records(self):
+        return self._test_handler.records
+
+    @property
+    def lines(self):
+        return self._test_handler.lines
+
+
+def test_logger():
+    return TestLogger()
```
atomatt: Is it worth keeping this now?

talisker/testing.py
```diff
@@ -0,0 +1,273 @@
...
+            strings_match = self._compare_strings(match, log)
+            extra_match = trailer_match = True
+
+            if extra is not None:
+                extra_match = self._compare_strings(extra, log['extra'])
+
+            if trailer is not None:
+                trailer_match = True
+                for line in trailer:
+                    if not any(line in tline for tline in log['trailer']):
```
atomatt: I wonder if the expected trailer lines should be in the correct order? I don't think it matters for now though.

talisker/testing.py
bloodearnest: I added it incase future code did return something in handle_exception, so that supression would work with in this odd 'chained' case. But I've reworked it now, so it's not needed.

talisker/testing.py
bloodearnest: yep

talisker/testing.py
bloodearnest: gone

PR #101, opened 2017-06-09
supress stderr on setup.py revision check

bloodearnest:
Also add missing test

### DIFF
--- a/talisker/revision.py
+++ b/talisker/revision.py
@@ -56,7 +56,8 @@ def bzr_version_info():
 
 
 def setup_py():
-    return subprocess.check_output([sys.executable, 'setup.py', '--version'])
+    return subprocess.check_output(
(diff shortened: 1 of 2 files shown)


talisker/revision.py
```diff
@@ -56,7 +56,8 @@ def bzr_version_info():
 
 
 def setup_py():
-    return subprocess.check_output([sys.executable, 'setup.py', '--version'])
+    return subprocess.check_output(
+        [sys.executable, 'setup.py', '--version'], stderr=subprocess.STDOUT)
```
cjwatson: Not sending to `/dev/null` instead?

bloodearnest: I want the output, to show in case of a test failure.

bloodearnest: Hmm, that was my thinking, but doesn't actuall make sense, as the output will be lost anyway.

PR #105, opened 2017-10-23
fix object encoding in logfmt

bloodearnest:


tests/test_logging.py
```diff
@@ -356,9 +356,12 @@ def test_logfmt_atom(monkeypatch):
...
     # quoting
-    assert fmt.logfmt_atom('foo', 'bar baz') == 'foo="bar baz"'
-    assert fmt.logfmt_atom('foo', 'bar\tbaz') == 'foo="bar\tbaz"'
-    assert fmt.logfmt_atom('foo', 'bar=baz') == r'foo="bar=baz"'
+    assert fmt.logfmt_atom('foo', 'string') == 'foo="string"'
+    assert fmt.logfmt_atom('foo', 1) == 'foo=1'
+    assert fmt.logfmt_atom('foo', True) == 'foo=true'
+    # str(d) is different in python 2/3, due to u'' prefix
+    d = {'foo': 'bar'}
+    assert fmt.logfmt_atom('foo', d) == 'foo="' + str(d) + '"'
```
sparkiegeek: hmm, what about dictionaries with keys (or values) that have a `"` embedded?

talisker/logs.py
```diff
@@ -351,6 +351,10 @@ def logfmt_atom(self, k, v):
                 v = v[:size] + self.TRUNCATED
 
             v = '"' + v + '"'
+        elif isinstance(v, bool):
+            v = str(v).lower()
+        elif not isinstance(v, int):
+            v = '"' + str(v).replace('"', '') + '"'
```
sparkiegeek: I think this would be better defined recursively

`return self.logfmt_atom(self, k, str(v))`

otherwise you risk hitting cases that you account for with the string handling e.g. blowing the size

talisker/logs.py
```diff
@@ -326,46 +327,67 @@ def format(self, record):
...
+        if isinstance(k, bytes):
+            k = k.decode('utf8')
 
-    def logfmt_atom(self, k, v):
-        if not v:
-            return None
+        # TODO: look at optimizing this to avoid string allocations
 
-        size = self.MAX_VALUE_SIZE
+        k = k[:self.MAX_KEY_SIZE].strip()
```
sparkiegeek: We should `.strip()` before truncating, so that leading whitespace won't cause the key to be unnecessarily truncated

PR #107, opened 2017-10-30
Handle more logfmt serialization corner cases and errors

bloodearnest:


tests/test_sentry.py
```diff
@@ -164,6 +164,7 @@ def app(environ, sr):
 
     mw = talisker.sentry.get_middleware(app)
     context = mw.client.context
+    context.clear()
```
bloodearnest: Note: this is related to getting the tests to pass when you have SENTRY_DSN set in the environment, seemed reasonable to include it.

talisker/logs.py
```diff
@@ -359,36 +365,63 @@ def logfmt_key(self, k):
...
+            v = '"' + str(type(v)) + '"'
 
         return v
 
+    def safe_string(self, s, max, truncate):
+        s = s.strip()
+
+        # no new lines allowed, so truncate at the first new line
+        if '\n' in s:
+            s = s.split('\n', 1)[0] + truncate
```
sparkiegeek: hmm, if the first line is longer than the maximum length then we'll add truncate twice!

bloodearnest:
Hah, good point.

talisker/logs.py
```diff
@@ -341,54 +349,100 @@ def logfmt_atoms(self, structured):
...
-        elif isinstance(v, dict):
-            return '"' + self.TRUNCATED + '"'
+        elif isinstance(v, numbers.Number):
+            v = str(v)
+        else:
+            v = '"' + str(type(v)) + '"'
 
         return v
 
+    def safe_string(self, s, max, truncate_str=None):
```
sparkiegeek: why is `truncate_str` optional? you're passing it through for both the key and value cases..

talisker/logs.py
```diff
@@ -341,54 +349,100 @@ def logfmt_atoms(self, structured):
...
+            truncated = True
+
+        # maximum size, to prevent overloading aggregator (accidental or
+        # malicious)
+        if len(s) > max:
+            s = s[:max]
+            truncated = True
+
+        # logstash kv filter cannot handle escaped ", so we just strip
+        s = s.replace('"', '')
```
sparkiegeek: there's a minor bug-let here where a string that is composed of one or more `"` and is above max-size could in fact be unnecessarily truncated (since we remove the `"` characters after testing for length)

talisker/logs.py
bloodearnest: Yes, and artifact for a previous iteration, will remove

talisker/logs.py
bloodearnest: hmm, right. I'm not overly concerned with that. The truncate length is an arbitrary limit to stop unbounded message size. A few chars +/- doesn't worry me. And doing it after truncate makes it faster, avoids needless replacements for things that are going to be discarded anyway.

sparkiegeek: :+1:

PR #114, opened 2018-02-12
Add a functional test suite

bloodearnest:
This PR adds a functional test suite, use the talisker.testing helpers added earlier.

I ended up avoiding pipes all together, to many issues with buffering and blocking. Using a temporary files allows for a nice writer fd and a reader fd to process the output without contention, and allows for a timeout by retrying the reader if it returns ''.


bloodearnest:
Splitting into other PRs

PR #118, opened 2018-02-16
Use file for subprocess output capture

bloodearnest:
This allows non-buffered and non-blocking reads from the standard out, which pipes did not allow reliably.

### DIFF
--- a/talisker/testing.py
+++ b/talisker/testing.py
@@ -21,15 +21,25 @@ from __future__ import division
 from __future__ import absolute_import
 
 from builtins import *  # noqa
-import sys
-import subprocess
+import logging
 import re
 import shlex
-import logging
+import subprocess
+import sys
+import tempfile
+import time
 
 import requests
 
 import talisker.logs
+
+
+if sys.version_info[0] == 2:
+    def temp_file():
+        return tempfile.NamedTemporaryFile('wb', bufsize=0)
+else:
+    def temp_file():
+        return tempfile.NamedTemporaryFile('wb', buffering=0)
 
 
 class TestHandler(logging.Handler):
@@ -76,11 +86,12 @@ class LogOutput:
     def read(self, lines):
         current = []
         for line in lines:
-            if self.TIMESTAMP.match(line):
-                if current:
-                    yield self.parse(current)
-                current = []
-            current.append(line)
+            if line.strip():
+                if self.TIMESTAMP.match(line):
+                    if current:
+                        yield self.parse(current)
+                    current = []
+                current.append(line)
         if current:
             yield self.parse(current)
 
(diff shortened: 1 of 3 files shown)


talisker/testing.py
```diff
@@ -21,17 +21,27 @@
...
 
 import talisker.logs
 
 
+if sys.version_info[0] == 2:
+    def temp_file():
+        return tempfile.NamedTemporaryFile('wb', bufsize=0)
+else:
+    def temp_file():
+        return tempfile.NamedTemporaryFile('wb', buffering=0)
```
cjwatson: Would it perhaps be better to pass `bufsize=1` / `buffering=1`?  That should result in line buffering, which I think should mean that you don't have to flush the output file in `readline` (modulo my question below about whether that flush actually does anything, but if you're reading line-by-line then line buffering still seems appropriate).

talisker/testing.py
```diff
@@ -140,12 +151,58 @@ class ServerProcessError(Exception):
...
+    def __init__(self, cmd, env=None):
         self.cmd = cmd
+        self.env = env
         self.output = []
         self.ps = None
         self._log = None
 
+    def start(self):
+        self.output_file = temp_file()
+        self.reader = open(self.output_file.name, 'r')
```
cjwatson: Is this guaranteed to be text and thus not to raise exceptions on Python 3 (answer may well be yes, but just checking)?

talisker/testing.py
```diff
@@ -140,12 +151,58 @@ class ServerProcessError(Exception):
...
+            raise
+
+    def close(self, error=False):
+        """Clean up process."""
+        if not self.finished:
+            self.ps.terminate()
+            self.ps.wait()
+
+        if self.output_file and not self.output_file.closed:
+            self.output_file.flush()
```
cjwatson: Does either this flush or the one in `readline` actually do anything?  This flushes the stream's write buffers, but it can't flush write buffers from some other process, and this file object is only ever written to by other processes.

talisker/testing.py
bloodearnest: This is tricky to get right I think, so I was cautious, as buffering doesn't provide much benefit here - we're expecting <100 of log lines for a unit test case.

The py2/py3 version makes it interesting too.

In py3, the file-like object I pass to Popen has to be in binary mode, and thus cannot be line buffered, only block buffered, which we don't really want.

In py2, it only supports block buffering anyway, AFAICS, or -1 for the operating system default (which is block buffered for stdout, right?)

So having no buffering for output, and using readline() for input, seemed the best from a correctness pov.

talisker/testing.py
bloodearnest: Yeah, the implicit assumption here is that we're testing against textual log output, as it's a helper for testing talisker programs. So I think the answer is yes :)

talisker/testing.py
bloodearnest: Right - the process is definitely dead by now, and thus any output flushed anyway. Removed.

PR #120, opened 2018-02-19
Add functional tests for gunicorn and celery

bloodearnest:


.travis.yml
```diff
@@ -32,4 +32,8 @@ addons:
   addons:
     postgresql: "9.5"
 install: pip install tox
+before_script:
+ - make db-setup
+services:
+    - redis-server
```
sparkiegeek: inconsistent indentation twitch

Makefile
```diff
@@ -50,13 +50,13 @@ _test: $(VENV)
 
 TEST_FILES = $(shell find tests -maxdepth 1 -name test_\*.py  | cut -c 7- | cut -d. -f1)
 $(TEST_FILES): $(VENV)
-	$(BIN)/pytest -k $@ $(ARGS)
+	. $(BIN)/activate && py.test -k $@ $(ARGS)
```
sparkiegeek: why this change? shouldn't pytest in the virtualenv already "apply" that virtualenv? maybe it's just the dependency that needs expressing?

Makefile
```diff
@@ -72,6 +72,11 @@ flask: | lib/sqlalchemy
 lib/redis:
 	$(BIN)/pip install redis
 
+db-setup:
+	psql -U postgres -c "create database django_app;"
+	psql -U postgres -c "create user django_app with password 'django_app';"
+	psql -U postgres -c "grant all privileges on database django_app to django_app;"
```
sparkiegeek: if you create the user first, then you could set the owner of the django_app db to the django_app user and skip the grant?

e.g. CREATE DATABASE django_app OWNER django_app

requirements.tests.txt
```diff
@@ -4,5 +4,9 @@ pytest-cov>=2.3.1
...
 pytest-xdist>=1.22.0
 setuptools>=35.0.0
-eventlet
-gevent
+
+# for integration tests
+eventlet==0.22.0
+gevent==1.2.2
+redis==2.10.6
+SQLAlchemy==1.1.14
```
sparkiegeek: is the strict version selector deliberate? What if downstream projects use different versions?

tests/test_functional.py
```diff
@@ -0,0 +1,85 @@
...
+from talisker.testing import GunicornProcess, ServerProcess
+from tests.celery_app import basic_task, error_task
+
+APP = 'tests.wsgi_app:application'
+
+
+def test_gunicorn_sync_worker():
+    with GunicornProcess(APP, args=['--worker-class=sync']) as p:
+        response = requests.get(p.url('/'))
+        assert response.status_code == 200
```
sparkiegeek: you could move the assert out of the context-manager?

tests/test_functional.py
```diff
@@ -0,0 +1,85 @@
...
+# You should have received a copy of the GNU General Public License
+# along with this program; if not, write to the Free Software
+# Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA
+
+from __future__ import unicode_literals
+from __future__ import print_function
+from __future__ import division
+from __future__ import absolute_import
+
+from builtins import *  # noqa
```
sparkiegeek: hmm? :confused: what's this for?

Makefile
bloodearnest: It's for the venv's $PATH.

The functional tests executes talisker.gunicorn and friends as subprocesses, which if we run the test suite with the venv activated, are on the $PATH. Otherwise we needed some messy path munging to guess the executable path. It's also in line with how tests are run by tox (i.e. in an activated venv), so standardises it.

Makefile
bloodearnest: Huh, TIL, nice

requirements.tests.txt
bloodearnest: This is for tests, so downstream won't use it, and is fixed in an upcoming PR anyway

tests/test_functional.py
bloodearnest: I could.

tests/test_functional.py
bloodearnest: It's to import the future backports in py2. In py3 it does nothing.

https://pypi.python.org/pypi/future

requirements.tests.txt
bloodearnest: Fixed as it just specifies a minimum

Makefile
sparkiegeek: :+1: thanks for the explanation

tests/test_functional.py
sparkiegeek: Ah, right. Having module names which don't match the package name always confuses me, but it's understandable in this case. Thanks!

PR #132, opened 2018-02-27
Requests metrics redux

bloodearnest:
Add support for customising metric names for http requests. Can now register a human-friendlt name for an ip, and also specify a number of path components to include in the metric name.

The implementation is a bit hackish, for 2 reasons:

1) we try to avoid requiring subclassing requests.Session and requests.TransportAdaptor, as these are frequently done in user code, and don't compose well. This means we decorate instance attributes at run time. Which does not taste good.

2) Ideally, we'd simply attach ``metric_path_len`` to the request object, and then use it in the response hook. But the request object is created and used directly inside the request function, there's no way we can do that unless we re-implement the entire request function, which could likely break lots of things.

I'd like to propose some changes to requests upstream to make this kind of customisation easier, if I find some time.

### DIFF
--- a/docs/requests.rst
+++ b/docs/requests.rst
@@ -36,7 +36,7 @@ the following metric name::
   my.prefix.requests.somehost-com.POST.200
 
 
-You can customised the name of this metric in a few ways. Firstly, if you are
+You can customise the name of this metric in a few ways. Firstly, if you are
 using IP addresses in your urls, you can give them a human friendly name for
 the metric::
 
@@ -48,7 +48,7 @@ Now, a request to http://1.2.3.4/ will emit a metric like so::
 
 Talisker does not include the url path in the metric name by default, as it
 could be highly variable, and thus create too many distinct metrics. But you
-can optionally allow a number of path compontents to be included in the
+can optionally allow a number of path components to be included in the
 name::
 
     session.get('https://somehost.com/some/url/XXX', metrics_path_len=2)
--- a/docs/requests.rst
+++ b/docs/requests.rst
@@ -8,8 +8,8 @@ Requests
 Enhanced session
 ----------------
 
-Talisker provides a way to upgrade a request.Session instance with two main
-extra features.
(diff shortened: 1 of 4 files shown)


talisker/requests.py
```diff
@@ -57,39 +69,66 @@ def configure(session):
...
-    prefix = 'requests.{}.{}.{}'.format(
-        parsed.hostname.replace('.', '-'),
+    if parsed.hostname in HOSTS:
+        hostname = HOSTS[parsed.hostname]
+    else:
+        hostname = parsed.hostname
+    if path_len > 0:
+        path_components = parsed.path.lstrip('/').split('/')
+        dotted_path = '.'.join(path_components[:path_len])
+        url = '{}.'.format(dotted_path)
```
cjwatson: This variable isn't really a URL, so a slightly different variable name might be less confusing for people skimming the code.  Maybe `url_components`?

docs/requests.rst
```diff
@@ -26,16 +26,38 @@ in the form:::
...
   session.post('https://somehost.com/some/url', data=...)
 
 would result in the duration of that request in ms being sent to statsd with
-the following metric name:::
+the following metric name::
 
   my.prefix.requests.somehost-com.POST.200
 
 
+You can customised the name of this metric in a few ways. Firstly, if you are
```
cjwatson: Typo for "customise".

docs/requests.rst
```diff
@@ -26,16 +26,38 @@ in the form:::
...
+
+    talisker.requests.register_ip('1.2.3.4', 'myhost')
+
+Now, a request to http://1.2.3.4/ will emit a metric like so::
+
+    my.prefix.requests.myhost.POST.200
+
+Talisker does not include the url path in the metric name by default, as it
+could be highly variable, and thus create too many distinct metrics. But you
+can optionally allow a number of path compontents to be included in the
```
cjwatson: Typo for "components".

PR #136, opened 2018-02-28
Request's breadcrumbs and logging

bloodearnest:
add breadcrumbs and logs to requests support, and allow disabling metrics for http requests.

### DIFF
--- a/talisker/requests.py
+++ b/talisker/requests.py
@@ -78,8 +78,8 @@ def configure(session):
     # If anyone has a better solution, *please* tell me!
     if not hasattr(session.send, '_inject_request_id'):
         session.send = inject_request_id(session.send)
-    if not hasattr(session.request, '_metric_path_len'):
-        session.request = enable_metric_path_len(session.request)
+    if not hasattr(session.request, '_metric_control'):
+        session.request = enable_metric_control(session.request)
 
 
 def inject_request_id(func):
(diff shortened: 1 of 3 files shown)


talisker/requests.py
```diff
@@ -88,33 +91,90 @@ def send(request, **kwargs):
         id = request_id.get()
         if id and HEADER not in request.headers:
             request.headers[HEADER] = id
-        return func(request, **kwargs)
+        try:
+            return func(request, **kwargs)
+        except Exception as e:
+            metadata = collect_metadata(request, None)
+            metadata['exception'] = str(e)
```
tomwardill: `str` or `repr` here?

talisker/requests.py
```diff
@@ -88,33 +91,90 @@ def send(request, **kwargs):
...
 
 
+def collect_metadata(request, response):
+    metadata = collections.OrderedDict()
+    metadata['url'] = request.url
+    metadata['method'] = request.method
+
+    if response is not None:
+        metadata['status_code'] = response.status_code
+        duration = response.elapsed.total_seconds() * 1000
```
tomwardill: seconds * 1000? What unit is this measuring in?

talisker/requests.py
```diff
@@ -88,33 +91,90 @@ def send(request, **kwargs):
...
+        # massage breadcrumb data to meet sentry expectations for http request
+        raven.breadcrumbs.record(
+            type='http', category='requests', data=metadata)
+
+        if getattr(_local, 'emit_metric', True):
+            path_len = getattr(_local, 'metric_path_len', 0)
+            metric_name = get_metric_name(response, path_len)
+            statsd.get_client().timing(metric_name, metadata['duration'])
+    except Exception:
+        logging.exception('response hook error')
```
tomwardill: Does this need to `raise` too?

talisker/requests.py
bloodearnest: ms

talisker/requests.py
bloodearnest: No, that's the point of adding the try, we don't want metrics failures to break user code

talisker/requests.py
bloodearnest: Hmm, perhaps, will change

PR #142, opened 2018-03-06
Fix some small logging errors

bloodearnest:


talisker/logs.py
```diff
@@ -176,7 +176,7 @@ def configure_test_logging():
     still add their own loggers to assert against log messages if needed.
     """
     set_logger_class()
-    handler = logging.NullHandler()
+    handler = logging.FileHandler('/dev/null')
```
tomwardill: Why this change? What is the difference between the `FileHandler` and the `NullHandler` in this case?

bloodearnest: This is to do with the breadcrumbs being generated by the *formatter*, not the *logger* (as we discussed in a hangout last week I think). Without it, no log breadcrumbs would be recorded during tests, as the NullHandler never emits any logs to the formatter.

Also, I think I prefer it as a default, as it means the tests fully execute the logging path every time. Given how much we mess with logging, I think that's a good change.

PR #151, opened 2018-03-07
add view name header

bloodearnest:
This will enable better metric names

### DIFF
--- a/talisker/flask.py
+++ b/talisker/flask.py
@@ -49,10 +49,22 @@ def sentry(app, dsn=None, transport=None, **kwargs):
     return sentry
 
 
+def add_view_name(response):
+    try:
+        endpoint = flask.request.endpoint
+        module = flask.current_app.view_functions[endpoint].__module__
+        response.headers['X-View-Name'] = module + '.' + endpoint
+    except Exception:
+        logging.getLogger(__name__).exception(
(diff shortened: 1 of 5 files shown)


talisker/flask.py
```diff
@@ -49,10 +49,22 @@ def sentry(app, dsn=None, transport=None, **kwargs):
     return sentry
 
 
+def add_view_name(response):
+    try:
+        endpoint = flask.request.endpoint
+        module = flask.current_app.view_functions[endpoint].__module__
+        response.headers['X-View-Name'] = module + '.' + endpoint
```
sparkiegeek: let's move this to a (new) `else:` clause, since it's not going to trigger an exception and is about handling the results of potential exception-raising code.

tests/test_flask.py
```diff
@@ -130,3 +130,15 @@ def test_register_app():
     assert tapp.config['LOGGER_HANDLER_POLICY'] == 'never'
     assert 'sentry' in tapp.extensions
     assert tapp.logger is logging.getLogger(tapp.logger_name)
+
+
+def test_flask_view_name_header():
+    tapp = Flask('test')
```
sparkiegeek: t'app for trouble at t'mill? :wink:

PR #164, opened 2018-03-19
Enhance exception support to collect errno/strerror from OSErrors

bloodearnest:


tests/py3_test_util.py
```diff
@@ -0,0 +1,76 @@
...
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program; if not, write to the Free Software
+# Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA
+
+# ignore this whole file for flake8, as when run underpy2 it will break
```
tomwardill: typo :)

tests/py3_test_util.py
```diff
@@ -0,0 +1,76 @@
...
+# flake8: noqa
+
+from __future__ import unicode_literals
+from __future__ import print_function
+from __future__ import division
+from __future__ import absolute_import
```
tomwardill: If we're definitely under py3, do we need these?

bloodearnest: hmm, maybe not

PR #171, opened 2018-03-27
Refactor requests metrics/logging, again

bloodearnest:
- remove previous attempts at controlling metric name, instead can just
 set metric_{api,host}_name kwargs
 - use view name from header if possible for metrics
 - better handling of custom dns names for metrics
 - redact querystring from logging/breadcrumbs
 - include errno fields in logging/breadcrumbs
 - metric for connection errors using new errno support

### DIFF
--- a/talisker/requests.py
+++ b/talisker/requests.py
@@ -134,7 +134,7 @@ def collect_metadata(request, response):
         ip = None
         netloc = parsed.netloc
 
-    # do not include querystring in url, as may have senstive
+    # do not include querystring in url, as may have senstive info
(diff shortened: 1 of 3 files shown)


talisker/requests.py
```diff
@@ -79,68 +80,94 @@ def configure(session):
...
+        ip = parsed.hostname
+        netloc = hostname
+        if parsed.port:
+            netloc += ':{}'.format(parsed.port)
+    else:
+        hostname = parsed.hostname
+        ip = None
+        netloc = parsed.netloc
+
+    # do not include querystring in url, as may have senstive
```
tomwardill: typo

PR #209, opened 2018-04-30
ci support flask 1.0

### LINKED ISSUE #208: API changes for the logger in Flask 1.0.0+
When upgrading to FLASK v1.0.0+ I get this error:

```bash
Traceback (most recent call last):
  File "PATH/lib/python3.6/site-packages/flask/_compat.py", line 35, in reraise
    raise value
  File "PATH/app.py", line 40, in <module>
    talisker.flask.register(app)
  File "PATH/lib/python3.6/site-packages/talisker/flask.py", line 62, in register
    app._logger = logging.getLogger(app.logger_name)
AttributeError: 'Flask' object has no attribute 'logger_name'
```

From [changelog](http://flask.pocoo.org/docs/1.0/changelog/):
> Flask.logger has been simplified. LOGGER_NAME and LOGGER_HANDLER_POLICY config was removed. The logger is always named flask.app. The level is only set on first access, it doesn’t check Flask.debug each time. Only one format is used, not different ones depending on Flask.debug. No handlers are removed, and a handler is only added if no handlers are already configured.

bloodearnest:
fixes #208


talisker/flask.py
```diff
@@ -52,36 +52,46 @@ def sentry(app, dsn=None, transport=None, **kwargs):
 def add_view_name(response):
     try:
         endpoint = flask.request.endpoint
-        module = flask.current_app.view_functions[endpoint].__module__
+        name = endpoint
+        if flask.current_app:
+            module = flask.current_app.view_functions[endpoint].__module__
```
tbille: When trying to access an endpoint that doesn't exist I get an Exception that is printed in the console:

```bash
2018-05-01 08:53:48.281Z ERROR talisker.flask "could not get view name from flask request" request_id="5007d585-1c42-49e0-a8f8-68fc7b4a8f0a"
Traceback (most recent call last):
  File "PATH/talisker/talisker/flask.py", line 57, in add_view_name
    module = flask.current_app.view_functions[endpoint].__module__
KeyError
```

bloodearnest: Ugh, that's ugly.

While its not directly related to this issue, I've pushed a fix to this branch for this issue, as it's a fairly small fix.

In the process, I discovered that raven.contrib.flask 5.27.0 error when there is no view, so will convert any 404 into a 500! So I moved the minimum supported version to the 5.27.1 bugfix release in the process.

PR #213, opened 2018-05-03
Add status pages for packages, workers, and objgraph

bloodearnest:


bloodearnest:
Example of workers page:
image

bloodearnest:
Example of packages page
image

bloodearnest:
Example of logtree page
image

bloodearnest:
Example of packages page
image

bloodearnest:
Example of objgraph page

bloodearnest:
Example of logtree page
image

talisker/endpoints.py
```diff
@@ -247,7 +260,221 @@ def metrics(self, request):
...
+def ul(items):
+    yield '<ul>'
+    for item in items:
+        yield '<li>{}</li>'.format(item)
+    yield '</ul>'
+
+
+def link(text, href, *args, **kwargs):
+    return '<a href="{}">{}</a>'.format(
+        href.format(*args, **kwargs),
```
cjwatson: Maybe `html.escape(href.format(*args, **kwargs), quote=True)`, just in case?

talisker/endpoints.py
```diff
@@ -21,15 +21,19 @@
 
 from builtins import *  # noqa
 
-import os
-import sys
 import collections
+from datetime import datetime
 import functools
+import html
```
cjwatson: Isn't this py3-only?  I thought talisker still wanted to work on py2.

talisker/__init__.py
```diff
@@ -87,6 +87,23 @@ def get_config(env=os.environ):
...
 
 
+TALISKER_ENV_VARS = {
+    # development
+    'DEVEL',
+    'DEBUGLOG',
+    'TALISKER_COLOR',
+    'TALISKER_LOGSTATUS'
+    # sentry config
+    'SENTRY_DSN',
```
cjwatson: On the first service I looked at (snapauth), `SENTRY_DSN` contains a username and password.  I can see how it's useful to log part of this, but you should strip authentication information from the URL.

talisker/endpoints.py
```diff
@@ -247,7 +260,221 @@ def metrics(self, request):
...
+def link(text, href, *args, **kwargs):
+    return '<a href="{}">{}</a>'.format(
+        href.format(*args, **kwargs),
+        html.escape(text.format(*args, **kwargs)),
+    )
+
+
+MASTER_FIELDS = [
+    'cwd',
+    'cmdline',
```
cjwatson: I guess there *shouldn't* be anything sensitive in the process command line, but it's not unheard of to perpetrate quick hacks there that then stick around.  Have you audited our existing services for sensitive information on command lines?

talisker/__init__.py
bloodearnest: Aha, right you are. I all ready have code to do that for logging sentry config, so I can reuse that. Good catch.

talisker/endpoints.py
bloodearnest: Talisker is using the future module, which provides a backport of py3's html module when running in py2.

talisker/__init__.py
bloodearnest: Hmm, do you think we should strip username and password, or just password?

talisker/endpoints.py
bloodearnest: So, for gunicorn, we do things via the charm layers, which don't use the cli to configure, we use config files. 

In theory, you could use --env  to pass secrets, or perhaps it could leak info about SSL config if you use that via the cli

bloodearnest: Hmm, I should check SCA

talisker/__init__.py
cjwatson: Just password would be fine.

docs/endpoints.rst
```diff
@@ -35,8 +35,20 @@ nagios checks.
...
-    Return some useful information about server status (TODO).
+``/_status/info/packages``
+    Shows a list of installed python packages and their versions
+
+``/_status/info/workers``
+    Shows a summary of master and worker processes (e.g CPU, memory, fd count)
+    and other process information.  *Only available if psutil is installed.*
+
+``/_status/info/objgraph``
+    Shows the most common python ojbects in user for the worker that services
```
sparkiegeek: s/ojbects/objects/

talisker/endpoints.py
```diff
@@ -120,35 +126,46 @@ def __init__(self, app, namespace='_status'):
...
+                continue
+            try:
                 func = getattr(self, funcname)
                 methods.append(
-                    item.format(self.prefix + url, url, func.__doc__))
-        return Response(
-            '<ul>' + '\n'.join(methods) + '<ul>', mimetype='text/html')
+                    link(url, self.prefix + url) + ' - ' + str(func.__doc__),
+                )
+            except AttributeError:
```
sparkiegeek: let's tighten up the `try`..`except` and move `methods.append()` to an `else` clause

talisker/endpoints.py
```diff
@@ -247,7 +260,222 @@ def metrics(self, request):
...
         import logging_tree
         tree = logging_tree.format.build_description()
         return Response(tree)
+
+    @private
+    def packages(self, request):
+        """List of python packages installed."""
+        import pkg_resources
+        pkgs = list(pkg_resources.working_set)
+        pkgs.sort(key=lambda p: p.project_name)
```
sparkiegeek: why do we explicitly convert to a list?

ISTM this just needs to be iterable?

i.e. `for p in sorted(pkg_resources.working_set, key=lambda p: p.project_name):`

talisker/endpoints.py
```diff
@@ -247,7 +260,222 @@ def metrics(self, request):
...
+    href="{cdn}/twitter-bootstrap/4.1.1/css/bootstrap.min.css">
+    <link rel="stylesheet"
+    href="{cdn}/bootstrap-table/1.12.1/bootstrap-table.min.css">
+    <script
+    src="{cdn}/jquery/3.3.1/jquery.js"></script>
+    <script
+    src="{cdn}/twitter-bootstrap/4.1.1/js/bootstrap.min.js"></script>
+    <script
+    src="{cdn}/bootstrap-table/1.12.1/bootstrap-table.min.js"></script>
+</head>
```
sparkiegeek: eek, that's a lot of dependencies

talisker/util.py
```diff
@@ -54,6 +54,25 @@ def parse_url(url, proto='http'):
...
+
+
+def sanitize_url(url):
+    """Strips a url of any pw and query strings."""
+    parsed = parse_url(url)
+    return CLEANED.format(
+        scheme=parsed.scheme,
+        user=parsed.username or '',
+        colon=':' if parsed.password else '',
+        password='*' * len(parsed.password) if parsed.password else '',
```
sparkiegeek: do we really want to leak len(password) ? I'd suggest hardcoding eight asterisks

talisker/endpoints.py
bloodearnest: I'd love a css only pretty table, but I couldn't find one.

talisker/util.py
bloodearnest: ok

PR #224, opened 2018-05-21
Make parsing the postgres dsn more robust

bloodearnest:


talisker/postgresql.py
```diff
@@ -56,11 +56,16 @@ def prettify_sql(sql):
...
+        try:
+            # 2.7+
+            params = conn.get_dsn_parameters()
+        except AttributeError:
+            params = dict(i.split('=') for i in shlex.split(conn.dsn))
+
+        params.setdefault('host', 'localhost')
+        return '{user}@{host}:{port}/{dbname}'.format(**params)
+    except Exception:
+        return None
```
tomwardill: Should this be something more helpful other than None? Maybe 'Unable to parse'?

bloodearnest: That's what I had originally. I supposed it surfaces the information better.

PR #231, opened 2018-05-31
add soft request timeout support

verterok:
This creates a sentry message when the duration of the request is longer than soft_request_timeout config value.

### DIFF
--- a/talisker/sentry.py
+++ b/talisker/sentry.py
@@ -25,6 +25,7 @@ import logging
 import os
 import time
 
+import flask
 import raven
 import raven.middleware
 import raven.handlers.logging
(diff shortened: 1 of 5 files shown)


verterok:
Looks like the failure is caused by a newer raven: 6.9.0, which includes: https://github.com/getsentry/raven-python/pull/1238

talisker/sentry.py
```diff
@@ -159,7 +161,28 @@ def __call__(self, environ, start_response):
...
+                if (soft_start_timeout is not None and
+                        duration > soft_start_timeout):
+                    self.client.captureMessage(
+                        'Start_response over timeout: {}'
+                        .format(soft_start_timeout)
+                    )
+                # only clear transaction via url_rule if there is a flask
+                # context
+                if flask.has_request_context() and flask.request.url_rule:
+                    self.client.transaction.pop(flask.request.url_rule.rule)
```
bloodearnest: Rather than pop it here, I think I'd prefer to clear the transaction completely at the start of the request, after the self.client.context.clear() on line 163 above, do a self.client.transaction.clear(). Then we don't need to worry about a flask context, and we preserve the current requests transactions stack for as long is possible, incase any other issue occur.

PR #232, opened 2018-05-31
fix failing test in tests/test_flask.py when using raven 6.9.0

verterok:
thought about making it backward compatible...but not sure there is a point in keep backward compatible tests


bloodearnest:
Sadly, we do need to keep it b/w compat, as we also test against the minimum version of raven we support in travis (currently 5.37 I think)

PR #233, opened 2018-05-31
Move the clearing of various request contexts as late as possible,

bloodearnest:
To basically just before another request starts.

This allows the request id and various sentry contexts to last as long
as possible, rather than being cleaned up at the wsgi stage. This means
gunicorn can have that info, and exceptions logged/reported (e.g. worker
timeouts) will have as much context as possible for logs and sentry
reports


verterok:
Looks good! thanks for taking care of the sentry context cleanup too.

PR #240, opened 2018-06-15
Fix prometheus multiprocess

bloodearnest:
prometheus_client needs the env var 'prometheus_multiproc_dir' set when it is imported, or else it imports in normal mode, not multiprocess mode. So we set this early. 

This does mean that we are always in multiprocess mode, even when we only have one worker, but I'm ok with that, as one worker is not an interesting use case for us.

There's also a driveby fix to get tests passing, as DNS config of the host made the test fail.

### DIFF
--- a/talisker/__init__.py
+++ b/talisker/__init__.py
@@ -27,7 +27,7 @@ import os
 import tempfile
 
 from future.utils import exec_
-from talisker.util import ensure_extra_versions_supported
+from talisker.util import ensure_extra_versions_supported, pkg_is_installed
 
 __version__ = '0.9.10'
 __all__ = [
@@ -46,7 +46,8 @@ def initialise(env=os.environ):
     # set this early so any imports of prometheus client will be imported
     # correctly
     if 'prometheus_multiproc_dir' not in os.environ:
-        os.environ['prometheus_multiproc_dir'] = tempfile.mkdtemp()
+        if pkg_is_installed('prometheus-client'):
+            os.environ['prometheus_multiproc_dir'] = tempfile.mkdtemp()
 
     import talisker.logs
     talisker.logs.configure(config)
--- a/talisker/__init__.py
+++ b/talisker/__init__.py
@@ -47,7 +47,8 @@ def initialise(env=os.environ):
     # correctly
     if 'prometheus_multiproc_dir' not in os.environ:
         if pkg_is_installed('prometheus-client'):
-            os.environ['prometheus_multiproc_dir'] = tempfile.mkdtemp()
+            tmp = tempfile.mkdtemp(prefix='prometheus_multiproc')
+            os.environ['prometheus_multiproc_dir'] = tmp
 
     import talisker.logs
     talisker.logs.configure(config)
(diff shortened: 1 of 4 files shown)


talisker/__init__.py
```diff
@@ -42,6 +43,11 @@
 
 def initialise(env=os.environ):
     config = get_config(env)
+    # set this early so any imports of prometheus client will be imported
+    # correctly
+    if 'prometheus_multiproc_dir' not in os.environ:
```
maxiberta: Maybe this should be preceded by a check `if pkg_is_installed('prometheus-client')`?

talisker/gunicorn.py
```diff
@@ -275,13 +274,9 @@ def load_config(self):
...
-            if 'prometheus_multiproc_dir' not in os.environ:
-                logger.info('running in multiprocess mode but '
-                            '`prometheus_multiproc_dir` envvar not set')
-                tmpdir = tempfile.mkdtemp()
-                os.environ['prometheus_multiproc_dir'] = tmpdir
-
-            logger.info('using `%s` for multiprocess prometheus metrics',
-                        os.environ['prometheus_multiproc_dir'])
+        dir = os.environ.get('prometheus_multiproc_dir')
+        if pkg_is_installed('prometheus-client') and dir is not None:
```
maxiberta: Can `dir` be `None` here? What if it's `None`?

talisker/gunicorn.py
```diff
@@ -275,13 +274,9 @@ def load_config(self):
...
-                (self.cfg.workers or 1) > 1):
-            if 'prometheus_multiproc_dir' not in os.environ:
-                logger.info('running in multiprocess mode but '
-                            '`prometheus_multiproc_dir` envvar not set')
-                tmpdir = tempfile.mkdtemp()
-                os.environ['prometheus_multiproc_dir'] = tmpdir
-
-            logger.info('using `%s` for multiprocess prometheus metrics',
-                        os.environ['prometheus_multiproc_dir'])
+        dir = os.environ.get('prometheus_multiproc_dir')
```
maxiberta: Beware `dir` is also a builtin function.

talisker/__init__.py
bloodearnest: Maybe, but there's no harm in setting the environment variable if prometheus-client isn't installed, it will do nothing.

talisker/gunicorn.py
bloodearnest: Good point

talisker/gunicorn.py
bloodearnest: right, it shouldn't be None ever, but defensively, it case it is None for some reason, don't log as the message as it's not true.

talisker/__init__.py
maxiberta: It's not just setting the envvar but creating a temp dir unnecessarily.

bloodearnest: Sure, but I don't have an issue with that particularly, but have added the protection.

talisker/__init__.py
```diff
@@ -42,6 +43,11 @@
...
+        os.environ['prometheus_multiproc_dir'] = tempfile.mkdtemp()
```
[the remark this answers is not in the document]
bloodearnest: Fixed

PR #256, opened 2018-07-20
- Do no swallow the Sentry capture's result.

### LINKED ISSUE #257: Do not swallow Sentry capture's result
As per the sentry doc https://docs.sentry.io/clients/python/api/, "capture" returns a "a tuple with a 32-length string identifying this event" which Talisker is not returning back in its SentryClient implementation.

nessita:
Fixes #257.
Simon, this needs a test but I'm proposing the change so this issue is raised to your attention.


bloodearnest:
Closing as is addressed in #259.

PR #262, opened 2018-07-26
Add support for TALISKER_REVISION_ID variable

### LINKED ISSUE #195: Ability to provide version-info.txt as environment variable?
It's useful to have the COMMIT_ID available to our application, and the best way for us to provide it is as an environment variable. But talisker reads it from the `version-info.txt` file instead. Would it be possible for talisker to pick it up from *either* the file *or* an environment variable? Just to save us having to [export it in 2 different ways](https://github.com/canonical-websites/snapcraft.io/blob/master/Dockerfile#L15).

nottrobin:
This can be used instead of `version-info.txt` or other methods of
determining the revision ID.

Fixes #195

### DIFF
--- a/talisker/revision.py
+++ b/talisker/revision.py
@@ -14,6 +14,7 @@
 # along with this program; if not, write to the Free Software
 # Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA
 
+import os
 import subprocess
 import sys
 
@@ -36,6 +37,10 @@ def _run(args):
 def version_info_txt():
     with open('version-info.txt', 'rb') as f:
         return f.read()
+
+
+def talisker_revision_id():
+    return os.environ.get('TALISKER_REVISION_ID').encode('utf-8')
 
 
 def git():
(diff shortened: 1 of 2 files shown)


tests/test_revision.py
```diff
@@ -22,6 +22,7 @@
 from builtins import *  # noqa
 
 from subprocess import check_output as run
+import os
```
bloodearnest: This isn't used, and so the linter is complaining

talisker/revision.py
```diff
@@ -38,6 +39,10 @@ def version_info_txt():
         return f.read()
 
 
+def talisker_revision_id():
+    return os.environ.get('TALISKER_REVISION_ID').encode('utf-8')
```
sparkiegeek: if there's no $TALISKER_REVISION_ID then won't this blow up?

`NoneType has no attribute 'encode'`

bloodearnest:
Yes it will, but it's wrapped in a try except when called, as protection
against revision functions (which can be user defined) blowing up in
general. Hence why tests passed.

That said, probably still cleaner to add an default param to get()

talisker/revision.py
nottrobin: Good point. I'll tidy that up.

nottrobin: Fixed in https://github.com/canonical-ols/talisker/pull/267 (I hope)

PR #267, opened 2018-07-30
Fix error case in talisker_revision_id

nottrobin:
As pointed out by sparkiegeek:
https://github.com/canonical-ols/talisker/pull/262#discussion_r206055685

### DIFF
--- a/talisker/revision.py
+++ b/talisker/revision.py
@@ -48,7 +48,9 @@ def version_info_txt():
 
 
 def talisker_revision_id():
-    return os.environ.get('TALISKER_REVISION_ID').encode('utf-8')
+    revision = os.environ.get('TALISKER_REVISION_ID')
+
+    return revision.encode('utf-8') if revision else None
 
 
 def git():


nottrobin:
@sparkiegeek we can't just `return os.environ.get('TALISKER_REVISION_ID')` (
I initially implemented it that way, and then had multiple failing tests.), because the processing function expects a `bytes` type, not a UTF-8 `str`. So `return os.environ.get('TALISKER_REVISION_ID')` works in Python 2, but not 3.

So I think this PR is needed to make the function not produce an error.

sparkiegeek:
@nottrobin ah, the `get()` does a `.decode()` and I misread it as a `.encode()` - ignore my previous comment.

nottrobin:
@sparkiegeek you merged it! I was gonna make it better!

sparkiegeek:
@nottrobin catching `AttributeError` does not count as 'better' :stuck_out_tongue:

PR #269, opened 2018-07-31
Make the get function more resilient

nottrobin:
So it can natively handle both python2 and python3 strings returned from
functions.


nottrobin:
@bloodearnest would you like me to just close this?

PR #276, opened 2018-08-15
Wrapping logging and metrics code with try/except, so that we are never

bloodearnest:
responsible for erroring the application if something unexpected
happens

### DIFF
--- a/tests/test_logging.py
+++ b/tests/test_logging.py
@@ -167,6 +167,18 @@ def test_make_record_ordering():
         'user1', 'user2', 'context', 'global']
 
 
+def test_make_record_protected(monkeypatch, log):
+    def error(*args):
+        raise Exception()
+
+    monkeypatch.setattr(logs.logging_context, 'get', error)
+    logger = logs.StructuredLogger('test')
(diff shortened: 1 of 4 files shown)


bloodearnest:
This pull request **introduces 1 alert** when merging ab0389b857a866306aed30eba740a4edc0455ca4 into 80bd3e564b45ea927997781da82710ee2873bb6b - [view on LGTM.com](https://lgtm.com/projects/g/canonical-ols/talisker/rev/pr-4e25c9527b5c72c7712e89ac20f9bdbaa4fa29b6)

**new alerts:**

* 1 for Wrong number of arguments in a call

---

*Comment posted by [LGTM.com](https://lgtm.com)*

tests/test_logging.py
```diff
@@ -276,6 +288,20 @@ def test_formatter_large_msg(monkeypatch):
     assert structured == {}
 
 
+def test_formatter_protected(monkeypatch):
+    fmt = logs.StructuredFormatter()
+
+    # make this formatter error
+    def error():
```
sparkiegeek: let's make this take an argument so it fails because of the `raise` rather than the wrong number of arguments

<3 LGTM for finding this.

PR #281, opened 2018-08-20
Aggregate prometheus metrics into one file on worker death.

bloodearnest:
This should stop the leaking files and subsequent increasing scrape
response times.

### DIFF
--- a/talisker/endpoints.py
+++ b/talisker/endpoints.py
@@ -123,6 +123,17 @@ def ok_response():
     return Response(str(talisker.revision.get()) + '\n')
 
 
+@module_cache
+def test_counter():
+    # this is created lazily as we may not have prometheus installed.
+    # Also, isolate in it's own registry. Multiprocess mode doesn't really use
+    # registries, but it helps keep things separated, which is useful in
+    # testing
+    import prometheus_client
+    registry = prometheus_client.CollectorRegistry()
+    return prometheus_client.Counter('test', 'test', registry=registry)
+
+
 class StandardEndpointMiddleware(object):
     """WSGI middleware to provide a standard set of endpoints for a service"""
 
@@ -251,10 +262,7 @@ class StandardEndpointMiddleware(object):
         if not pkg_is_installed('prometheus-client'):
             return Response('Not Supported', status=501)
 
-        if not hasattr(self, 'test_counter'):
-            import prometheus_client
-            self.test_counter = prometheus_client.Counter('test', 'test')
-        self.test_counter.inc()
+        test_counter().inc()
         return Response('Incremented test counter')
 
     @private
--- a/talisker/endpoints.py
+++ b/talisker/endpoints.py
@@ -126,9 +126,9 @@ def ok_response():
 @module_cache
 def test_counter():
     # this is created lazily as we may not have prometheus installed.
-    # Also, isolate in it's own registry. Multiprocess mode doesn't really use
+    # Also, isolate in its own registry. Multiprocess mode doesn't really use
     # registries, but it helps keep things separated, which is useful in
-    # testing
+    # testing.
     import prometheus_client
     registry = prometheus_client.CollectorRegistry()
     return prometheus_client.Counter('test', 'test', registry=registry)
--- a/talisker/endpoints.py
+++ b/talisker/endpoints.py
@@ -129,9 +129,8 @@ def test_counter():
     # Also, isolate in its own registry. Multiprocess mode doesn't really use
     # registries, but it helps keep things separated, which is useful in
     # testing.
-    import prometheus_client
-    registry = prometheus_client.CollectorRegistry()
-    return prometheus_client.Counter('test', 'test', registry=registry)
+    from prometheus_client import Counter, CollectorRegistry
+    return Counter('test', 'test', registry=CollectorRegistry())
 
 
 class StandardEndpointMiddleware(object):
--- a/talisker/metrics.py
+++ b/talisker/metrics.py
@@ -303,14 +303,18 @@ def write_metrics(metrics, histogram_file, counter_file):
     histograms = _MmapedDict(histogram_file)
     counters = _MmapedDict(counter_file)
 
-    for metric in metrics:
-        if metric.type == 'histogram':
-            sink = histograms
-        elif metric.type == 'counter':
-            sink = counters
+    try:
+        for metric in metrics:
+            if metric.type == 'histogram':
+                sink = histograms
+            elif metric.type == 'counter':
+                sink = counters
 
-        for name, labels, value in metric.samples:
-            key = json.dumps(
-                (metric.name, name, tuple(labels), tuple(labels.values()))
-            )
-            sink.write_value(key, value)
+            for name, labels, value in metric.samples:
+                key = json.dumps(
+                    (metric.name, name, tuple(labels), tuple(labels.values()))
+                )
+                sink.write_value(key, value)
+    finally:
+        histograms.close()
+        counters.close()
--- a/talisker/metrics.py
+++ b/talisker/metrics.py
@@ -31,12 +31,17 @@ from builtins import *  # noqa
 
 import functools
 import logging
+import json
+import os
+import shutil
+import tempfile
 
 import talisker.statsd
 
 
 try:
     import prometheus_client
+    from prometheus_client import multiprocess
 except ImportError:
     prometheus_client = False
 
@@ -44,6 +49,8 @@ try:
     import statsd
 except ImportError:
     statsd = False
+
+logger = logging.getLogger(__name__)
 
 
 logger = logging.getLogger(__name__)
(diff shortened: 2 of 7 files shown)


talisker/metrics.py
```diff
@@ -112,3 +121,126 @@ def inc(self, amount=1, **labels):
...
+
+    all_files = [histogram_archive, counter_archive] + list(files)
+    tmp = tempfile.mkdtemp(prefix='prometheus_aggregate')
+    try:
+        for filename in all_files:
+            path = os.path.join(prom_dir, filename)
+            if os.path.exists(path):
+                shutil.copy(path, tmp)
+
+        collector = multiprocess.MultiProcessCollector(CollectorRegistry())
```
maxiberta: Shouldn't `collector` receive the `tmp` path as a parameter? Or maybe I'm missing something.

tests/conftest.py
```diff
@@ -29,16 +29,22 @@
 
 from builtins import *  # noqa
 
+# make sure prometheus is setup in multiprocess mode. We don't actuall use this
+# dir in tests, as each test gets it's own directory, but this ensures
```
maxiberta: s/it's/its/.

talisker/endpoints.py
```diff
@@ -123,6 +123,17 @@ def ok_response():
     return Response(str(talisker.revision.get()) + '\n')
 
 
+@module_cache
+def test_counter():
+    # this is created lazily as we may not have prometheus installed.
+    # Also, isolate in it's own registry. Multiprocess mode doesn't really use
+    # registries, but it helps keep things separated, which is useful in
+    # testing
```
sparkiegeek: missing full stop

talisker/metrics.py
```diff
@@ -132,3 +139,175 @@ def inc(self, amount=1, **labels):
...
+
+def histogram_sorter(sample):
+    # sort histogram samples in order of bucket size
+    name, labels, _ = sample
+    return name, float(labels.get('le', 0))
+
+
+def prometheus_cleanup_worker(pid):
+    """Clean up after a multiprocess worker has died."""
+    if prometheus_client is None:
```
sparkiegeek: what's this guarding against? If prometheus_client fails to import it's set to `False`, not `None`

talisker/metrics.py
```diff
@@ -132,3 +139,175 @@ def inc(self, amount=1, **labels):
...
+            tmp_histogram.name, os.path.join(prom_dir, histogram_archive))
+        os.rename(
+            tmp_counter.name, os.path.join(prom_dir, counter_archive))
+
+        for path in paths:
+            if os.path.exists(path):
+                os.unlink(path)
+
+    except Exception:
+        # we should never fail at cleaning up
```
sparkiegeek: agree with the sentiment, let's put it in the caller though? i.e. `prometheus_multiprocess_worker_exit`

talisker/metrics.py
```diff
@@ -132,3 +139,175 @@ def inc(self, amount=1, **labels):
...
+        prom_dir = os.environ['prometheus_multiproc_dir']
+        worker_files = [
+            'histogram_{}.db'.format(pid),
+            'counter_{}.db'.format(pid),
+        ]
+        paths = [os.path.join(prom_dir, f) for f in worker_files]
+
+        # check at least one worker file exists
+        if not any(os.path.exists(path) for path in paths):
+            return
```
sparkiegeek: how about putting the `os.path.exists()` check in the list-comp for `paths`?

That way we can just check `if not paths` here, and drop the check on `unlink` too?

talisker/metrics.py
```diff
@@ -132,3 +139,175 @@ def inc(self, amount=1, **labels):
...
+                    buckets[without_le][bucket[0]] += value
+                else:
+                    # _sum/_count
+                    samples[(name, labels)] += value
+
+            else:
+                # Counter and Summary.
+                samples[(name, labels)] += value
+
+        # modified to remove accumulation
```
sparkiegeek: let's explicitly mark the EOVERBATIM?

talisker/metrics.py
```diff
@@ -132,3 +139,175 @@ def inc(self, amount=1, **labels):
...
+
+    1. it takes its files as an argument, rather than hardcoding '*.db', an
+       skips none existent files
+    2. it does not accumulate histograms, as it is intended for writing that
+       data back out
+    3. it uses an OrderedDict to preserve label order, and thus metric identity
+
+    It needs to be kept up to date with changes to prometheus_client as much as
+    possible, or until changes are landed upstream to allow better reuse.
+    """
```
sparkiegeek: WDYT about an explicit whitelist and failure if we're using a version of prometheus_client that's not been checked?

It means we have to explicitly double check this when we bump prometheus_client but I think that's a Good Thing:tm:

talisker/metrics.py
```diff
@@ -132,3 +139,175 @@ def inc(self, amount=1, **labels):
...
+        # we should never fail at cleaning up
+        logger.exception('failed to cleanup prometheus worker files')
+
+
+def collect(files):
+    """This almost verbatim from MultiProcessCollector.collect().
+
+    It differs in a few ways:
+
+    1. it takes its files as an argument, rather than hardcoding '*.db', an
```
sparkiegeek: s/an/and/

tests/conftest.py
```diff
@@ -29,16 +29,22 @@
 
 from builtins import *  # noqa
 
+# make sure prometheus is setup in multiprocess mode. We don't actuall use this
```
sparkiegeek: *actually

tests/test_metrics.py
```diff
@@ -29,69 +29,83 @@
...
         statsd='{name}.{label}',
+        registry=registry,
     )
 
     labels = {'label': 'value'}
-    count = get_metric('test_histogram_count', **labels)
-    sum = get_metric('test_histogram_sum', **labels)
-    bucket = get_metric('test_histogram_bucket', le='2.5', **labels)
+    count = registry.get_metric('test_histogram_count', **labels)
+    sum = registry.get_metric('test_histogram_sum', **labels)
```
sparkiegeek: let's not shadow the builtin?

talisker/metrics.py
bloodearnest: It's just defence in depth, but I've fixed the check

talisker/metrics.py
bloodearnest: Well, then I have to join twice.
If could do it in two steps, though, will try that

talisker/metrics.py
```diff
@@ -45,6 +50,8 @@
 except ImportError:
     statsd = False
 
+logger = logging.getLogger(__name__)
```
maxiberta: This looks like a duplicate line.

talisker/metrics.py
bloodearnest: I think I would probably prefer to just issue a warning on startup. We can't control what versions of prometheus_client is used in the main project, as I wouldn't want to error or block if they it goes beyond what talisker supports.

maxiberta: This code duplication looks dangerous to me; I'd rather fork `prometheus_client` instead. That said, can't why "unaccumulate" results as in previous versions of this branch?

bloodearnest: Once we've validated this approach in prod, I intend to propose some small changes upstream that will mean this duplication can disappear, time permitting.

bloodearnest: @maxiberta we can unaccumlate, and I did, but we cannot preserve label order without duplicating, and that is essential.

I don't think forking the whole library is a better solution that a custom collection function, that is only used for one specific purpose.

bloodearnest: If you look at the first commit, you can see an approach that did not duplicate. I got it 90% of the way there, only to fail at the final hurdle.

bloodearnest: ultimately, prometheus scraping is not sensitive to label order, but the mmaped file format parsing is

bloodearnest:
This pull request **introduces 2 alerts** and **fixes 1** when merging 3e684dc20a5f77ba982c6303cc361156068d078d into dabd40b8b4a5a2a630661cbad960ca18bdfb5143 - [view on LGTM.com](https://lgtm.com/projects/g/canonical-ols/talisker/rev/pr-eb4a308134d937232dc45207917b9d51f83fb874)

**new alerts:**

* 2 for Module is imported with &#39;import&#39; and &#39;import from&#39;

**fixed alerts:**

* 1 for Module is imported with &#39;import&#39; and &#39;import from&#39;

---

*Comment posted by [LGTM.com](https://lgtm.com)*

bloodearnest:
This pull request **introduces 2 alerts** and **fixes 1** when merging c14f4e3cef492cadfebd2b94a8539511d798a0eb into dabd40b8b4a5a2a630661cbad960ca18bdfb5143 - [view on LGTM.com](https://lgtm.com/projects/g/canonical-ols/talisker/rev/pr-60c33609e555175a76e0cf345ae2fcc19569a7d6)

**new alerts:**

* 2 for Module is imported with &#39;import&#39; and &#39;import from&#39;

**fixed alerts:**

* 1 for Module is imported with &#39;import&#39; and &#39;import from&#39;

---

*Comment posted by [LGTM.com](https://lgtm.com)*

talisker/metrics.py
```diff
@@ -132,3 +137,180 @@ def inc(self, amount=1, **labels):
...
+        if metric.type == 'histogram':
+            sink = histograms
+        elif metric.type == 'counter':
+            sink = counters
+
+        for name, labels, value in metric.samples:
+            key = json.dumps(
+                (metric.name, name, tuple(labels), tuple(labels.values()))
+            )
+            sink.write_value(key, value)
```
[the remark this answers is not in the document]
bloodearnest: hmm, probably, although there various fd's should be closed when they go out of scope.

But will close them explicitly

bloodearnest:
This pull request **introduces 1 alert** and **fixes 1** when merging 5cfa09854d0b6bf8c6e48caf2998c0a5b2b267f4 into dabd40b8b4a5a2a630661cbad960ca18bdfb5143 - [view on LGTM.com](https://lgtm.com/projects/g/canonical-ols/talisker/rev/pr-322a207ae3261226af2383ce417a1b78074ae3df)

**new alerts:**

* 1 for Module is imported with &#39;import&#39; and &#39;import from&#39;

**fixed alerts:**

* 1 for Module is imported with &#39;import&#39; and &#39;import from&#39;

---

*Comment posted by [LGTM.com](https://lgtm.com)*

Issue #287, opened 2018-08-21
Prometheus opens file with master pid


bloodearnest:
Due to prometheus_client implementation, any metrics created early will use the master pid in it's mmaped filename. The first time a metric is written, this will be fixed, as it checks for pid changes on writing.

This means there might be an empty {histogram,counter}_$MASTERPID.db file left around for the duration of the gunicorn master.

Not a huge issue, but would be nice to fix at some point, but is not easy, given prometheus_client implementation choices.

PR #288, opened 2018-08-21
Fix issue with paths calculations which led to empty mmaped files on disk

bloodearnest:


maxiberta:
No stale .db files, except a "counter_<master PID>.db" file (but no "histogram_<master PID>.db"); is that expected?

maxiberta:
Ok, it's a known issue (https://github.com/canonical-ols/talisker/issues/287). +1

PR #296, opened 2018-08-28
Properly serialize the prometheus worker cleanup

bloodearnest:
The child_exit hook now just records the pid that needs cleaning, the
actual work is done in gunicorn's main loop, serialized and safe. To do
this we have to add a fake 'signal' to hook into gunicorns loop, but it
works.

In the process we also drop support for gunicorn <19.7, as they don't have a child_exit hook.

### DIFF
--- a/talisker/gunicorn.py
+++ b/talisker/gunicorn.py
@@ -29,11 +29,10 @@ from __future__ import absolute_import
 
 from builtins import *  # noqa
 
-from collections import OrderedDict
+from collections import OrderedDict, deque
 import logging
 import os
 
-import gunicorn.config
 from gunicorn.glogging import Logger
 from gunicorn.app.wsgiapp import WSGIApplication
 
@@ -59,6 +58,9 @@ DEVEL_SETTINGS = {
 }
 
 
+logger = logging.getLogger(__name__)
+
+
 class GunicornMetric:
     latency = talisker.metrics.Histogram(
         name='gunicorn_latency',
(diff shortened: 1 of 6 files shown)


talisker/metrics.py
```diff
@@ -149,45 +153,36 @@ def histogram_sorter(sample):
...
 
-        os.rename(tmp_histogram.name, histogram_path)
-        os.rename(tmp_counter.name, counter_path)
+    tmp_histogram = tempfile.NamedTemporaryFile(delete=False)
+    tmp_counter = tempfile.NamedTemporaryFile(delete=False)
+    write_metrics(metrics, tmp_histogram.name, tmp_counter.name)
 
-        for path in paths:
-            os.unlink(path)
+    # TODO: race condition window starts here
```
maxiberta: This'd be a new worker reusing the same PID, right?

talisker/gunicorn.py
```diff
@@ -221,24 +278,8 @@ def init(self, parser, opts, args):
...
-
-            def prometheus_multiprocess_worker_exit(server, worker):
-                "Worker cleanup function for multiprocess prometheus_client."
-                logging.getLogger(__name__).info(
-                    'Performing multiprocess prometheus_client cleanup')
-                cleanup(worker.pid)
-
-            cfg[server_hook] = prometheus_multiprocess_worker_exit
+            cfg['on_starting'] = gunicorn_on_starting
+            cfg['child_exit'] = gunicorn_child_exit
```
maxiberta: This is the default behaviour in talisker, but can still be completely overriden by users via standard gunicorn config, which is the right thing to do to me. Am I right?

talisker/gunicorn.py
```diff
@@ -82,8 +84,63 @@ class GunicornMetric:
     )
 
 
+# We add a synthetic signal, SIGCUSTOM, to gunicorn's known signals. This
+# allows us to get gunicorn to process the effects of this signal in the
+# arbiter's main loop, rather than within the limited context of the signal
+# handlers.  This makes worker clean up serialized and normal python code.
+
+DEAD_WORKERS = deque()  # storage for recording dead pids
```
maxiberta: Smart use of `deque`!

talisker/gunicorn.py
```diff
@@ -82,8 +84,63 @@ class GunicornMetric:
...
+                extra={'pid': pid},
+            )
+
+
+def gunicorn_on_starting(arbiter):
+    """Gunicorn on_starging server hook.
+
+    Sets up the fake signal and handler on the arbiter instance.
+    """
+    arbiter.SIG_NAMES['SIGCUSTOM'] = 'custom'
```
maxiberta: `SIGCUSTOM`, `handle_custom()`, etc sound a bit too generic to me. But I don't have a better alternative so I'll shut up :-)

bloodearnest: Yeah, I left it generic as it's just a generic way to hook into gunicorn's default main loop signal processing. We might very well use it for something other than prometheus clean up in the future.

SIGTALISKER maybe?

It's doesn't actually need a SIG prefix, and string will do, but I felt it should conform.

talisker/gunicorn.py
bloodearnest: Yes, that's right. If you override too many things yourself though stuff may break.

talisker/metrics.py
bloodearnest: No, that's a slightly different issue. 

This is:
master: read counter_<pid>.db, and write updated counter_archive file
worker, handling scrape concurrently: read updated counter_archive file, as well as not yet deleted counter_<pid>.db file, and double count them.
master: delete counter_<pid>.db.

It's a small window, but I have seen it in real life. We need a way to make the rename/delete atomic between multiple processes, which probably means a file-system based lock. But it's a separate issue.

bloodearnest:
I will add an issue with explanation, for posterity.

bloodearnest:
I added a multiprocessing.Lock() around the update phrase. I don't know how to test this, but it doesn't break anything!

bloodearnest:
Added an issue for posterity, this branch should fix #297

PR #304, opened 2018-09-11
Improve the ip->name mapping

bloodearnest:
Now include port, and is more explicit about net location in logging metadata

### DIFF
--- a/talisker/requests.py
+++ b/talisker/requests.py
@@ -33,6 +33,7 @@ import collections
 import functools
 import logging
 import threading
+import warnings
 
 from future.moves.urllib.parse import parse_qsl
 from future.utils import text_to_native_str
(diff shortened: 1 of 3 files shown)


talisker/requests.py
```diff
@@ -222,6 +230,7 @@ def metrics_response_hook(response, **kwargs):
 
 
 def record_request(request, response=None, exc=None):
+    import pdb; pdb.set_trace()
```
sparkiegeek: err...

talisker/requests.py
```diff
@@ -85,9 +85,19 @@ class RequestsMetric:
...
-    """Register a human friendly name for an IP address for metrics."""
-    HOSTS[ip] = name
+def register_endpoint_name(endpoint, name):
+    """Register a human friendly name for an IP:PORT address for metrics."""
+    parsed = parse_url(endpoint)
+    HOSTS[parsed.netloc] = name
+
+
+# backwards compat alias
+register_ip = register_endpoint_name
```
sparkiegeek: can we wrap this in a decorator like https://python-deprecated.readthedocs.io/en/latest/tutorial.html ?

talisker/requests.py
```diff
@@ -85,9 +85,19 @@ class RequestsMetric:
...
+
+
+def get_endpoint_name(endpoint):
+    parsed = parse_url(endpoint)
+    return HOSTS.get(parsed.netloc, None)
```
sparkiegeek: the second argument here is useless, `.get()` will do that out of the box.

PR #305, opened 2018-09-12
add explicit context clears at various points

bloodearnest:


talisker/gunicorn.py
```diff
@@ -110,6 +110,11 @@ def handle_custom():
                 'failed to cleanup prometheus worker files',
                 extra={'pid': pid},
             )
+    # Clear any sentry breadcrumbs that might have built up
+    # This is not ideal, but it's the only place the master process calls our
+    # code, so...
+    # If we don't, breadcrumbs keep on collecting for ever.
```
sparkiegeek: s/for ever/forever/

PR #334, opened 2018-10-10
Support prometheus 0.4+

bloodearnest:
Prometheus 0.4+ has a number of changes that affect talisker's integration

1) there is now a merge method, which is good, as we added that, so we can use it :)
2) the format of samples changed from a 3-tuple to a 5-tuple
3) the name of the counter metric changed, it now adds '_total' on the end.

Hopefully we can drop support for prometheus_client <0.4 soonish, to avoid some of the abstraction added here.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -46,7 +46,7 @@ lint: $(VENV)
 	$(BIN)/flake8 talisker tests
 
 _test: $(VENV)
-	. $(BIN)/activate && $(BIN)/pytest --tb=short -n auto $(ARGS)
+	. $(BIN)/activate && $(BIN)/pytest --timeout=15 $(ARGS)
 
 TEST_FILES = $(shell find tests -maxdepth 1 -name test_\*.py  | cut -c 7- | cut -d. -f1)
 $(TEST_FILES): $(VENV)
--- a/Makefile
+++ b/Makefile
@@ -46,7 +46,7 @@ lint: $(VENV)
 	$(BIN)/flake8 talisker tests
 
 _test: $(VENV)
-	. $(BIN)/activate && $(BIN)/pytest --timeout=15 $(ARGS)
+	. $(BIN)/activate && $(BIN)/pytest -n auto --timeout=15 $(ARGS)
 
 TEST_FILES = $(shell find tests -maxdepth 1 -name test_\*.py  | cut -c 7- | cut -d. -f1)
 $(TEST_FILES): $(VENV)
--- a/talisker/metrics.py
+++ b/talisker/metrics.py
@@ -170,9 +170,11 @@ def histogram_sorter(sample):
 
 
 def _filter_exists(paths):
+    exists = []
     for path in paths:
-        if os.path.exists(paths):
-            yield path
+        if os.path.exists(path):
+            exists.append(path)
+    return exists
 
 
 def prometheus_cleanup_worker(pid):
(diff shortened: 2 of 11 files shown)


Makefile
```diff
@@ -46,7 +46,7 @@ lint: $(VENV)
 	$(BIN)/flake8 talisker tests
 
 _test: $(VENV)
-	. $(BIN)/activate && $(BIN)/pytest --tb=short -n auto $(ARGS)
+	. $(BIN)/activate && $(BIN)/pytest --timeout=15 $(ARGS)
```
sparkiegeek: Huh?

talisker/metrics.py
```diff
@@ -183,8 +192,15 @@ def prometheus_cleanup_worker(pid):
 
     histogram_path = os.path.join(prom_dir, histogram_archive)
     counter_path = os.path.join(prom_dir, counter_archive)
+    archive_paths = [histogram_path, counter_path]
+    archive_paths = [p for p in archive_paths if os.path.exists(p)]
```
sparkiegeek: Why not `_filter_exists`?

talisker/metrics.py
```diff
@@ -183,8 +192,15 @@ def prometheus_cleanup_worker(pid):
...
 
-    metrics = collect(paths + [histogram_path, counter_path])
+    collect_paths = paths + archive_paths
+    try:
+        collector = MultiProcessCollector(None)
```
sparkiegeek: pull this out of the `try..except..` since we don't expect it to raise `AttributeError`

tox.ini
```diff
@@ -7,7 +7,7 @@ skipsdist = True
 passenv = TRAVIS TRAVIS_JOB_ID TRAVIS_BRANCH
 usedevelop = True
 deps = -r{toxinidir}/requirements.tests.txt
-commands = py.test -n auto --cov=talisker
+commands = py.test --cov=talisker
```
sparkiegeek: Double "huh?"

Makefile
bloodearnest: Yeah, something is broken with the tests in pytest-xdist. I was regularly getting test hangs in parallel, but not when run normally.Somehow the way I altered the prometheus test made it start happening reliably.

I need to get to the bottom of it, but it's tricky to debug, so I disabled parallel tests for now. I hope to be able to investigate and renable, but I didn't want to block on it.

talisker/metrics.py
bloodearnest: oops, yes

talisker/metrics.py
bloodearnest: ok

tox.ini
bloodearnest: See above. Travis is already more susceptible to these weird test hangs.

talisker/metrics.py
```diff
@@ -40,11 +40,25 @@
...
+        mark_process_dead,
+    )
+    prom_version = pkg_version('prometheus_client')
+    if prom_version in ('0.4.0', '0.4.1'):
+        raise TaliskerVersionException(
+            'prometheus_client {} has a critical bug in multiprocess mode, '
+            'and is not supported in Talisker. '
+            'https://github.com/prometheus/client_python/issues/322'.format(
+                prom_version,
+            )
```
sparkiegeek: the `.format` doesn't work like that with the magic string concatenation, you'll either have to wrap the template in ()s, or move the `.format()` on the line with the `{}` but then you don't get the magic concatenation

bloodearnest: Narrator: Adam was wrong.

;P

PR #344, opened 2018-10-23
Refactor the way Talisker keeps sentry client references up to date.

bloodearnest:
Move to a property based lookup, rather than tracking/updating actual
references.  This enables a future branch that changes how testing
sentry works.

This is needed because flask/django create/configure new sentry client
instances, using their relative app configuration, when they are
initialised. Talisker has by this point already set up various handlers
(e.g. logging, WSGI), and these have a reference to the client they were
created with (which may be missing application configuration). So we
need to update those to point to the new client instance. This is done
by subclassing and using a property to make the .client attribute
actually be a call to get_client() under the hood.


talisker/sentry.py
```diff
@@ -137,7 +126,12 @@ def log_client(client, from_env=False):
         return
 
     # base_url shouldn't have secrets in, but just in case, clean it
-    public_dsn = client.remote.get_public_dsn()
+    try:
+        public_dsn = client.remote.get_public_dsn()
+    except Exception:
+        # people patch is_enabled in tests, don't fail here if there is no DSN
```
jdahlin: This is usually too broad, it'll catch AttributeError, TypeError etc. Is it possible to catch only the case when there's no DSN set? Perhaps by looking at the configuration?

talisker/flask.py
```diff
@@ -44,6 +44,22 @@
...
+    def client(self):
+        """Return None if not yet set, so we do actually create the client."""
+        if self._client_set:
+            return talisker.sentry.get_client()
+        else:
+            return None
+
+    @client.setter
+    def client(self, client):
+        """We let the flask extention create the sentry client."""
```
sparkiegeek:
```suggestion
        """We let the flask extension create the sentry client."""
```

talisker/sentry.py
```diff
@@ -210,15 +199,44 @@ def __init__(self, *args, **kwargs):
...
         data = super().build_msg(event_type, *args, **kwargs)
         add_talisker_context(data)
         return data
 
 
-class TaliskerSentryMiddleware(raven.middleware.Sentry):
+class ProxyClientMixin(object):
+    """Mixin that overrides self.client to be property.
+
+    This allows use to reuse the various raven handlers, which use a client
```
sparkiegeek:
```suggestion
    This allows reuse of the various raven handlers, which use a client
```

talisker/sentry.py
```diff
@@ -210,15 +199,44 @@ def __init__(self, *args, **kwargs):
...
 
     def build_msg(self, event_type, *args, **kwargs):
         data = super().build_msg(event_type, *args, **kwargs)
         add_talisker_context(data)
         return data
 
 
-class TaliskerSentryMiddleware(raven.middleware.Sentry):
+class ProxyClientMixin(object):
+    """Mixin that overrides self.client to be property.
```
sparkiegeek:
```suggestion
    """Mixin that overrides self.client to be a property.
```

talisker/sentry.py
```diff
@@ -210,15 +199,44 @@ def __init__(self, *args, **kwargs):
...
+    This allows use to reuse the various raven handlers, which use a client
+    attribute, but have that attribute actually get the global client."""
+
+    @property
+    def client(self):
+        return get_client()
+
+    @client.setter
+    def client(self, client):
+        "Ignore, as we should be using the global client."""
```
sparkiegeek:
```suggestion
        """Ignore, as we should be using the global client."""
```

PR #386, opened 2018-12-17
Add documentation for config, including talisker.help command

bloodearnest:


bloodearnest:
The output of talisker.help currently looks like this: 

```
Usage: talisker.help [CONFIG NAME]

Talisker provides some executable wrappers, which initialise Talisker
and then simply pass through any supplied arguments to underlying command.

 - talisker.gunicorn wraps the regular gunicorn invocation
 - talisker.celery wraps the celery command
 - talisker.run wraps a regular call to python, and takes a script to run

Talisker can be configured by the environment variables listed below. These
variable can also be supplied in a python file, although environment
variables override any file configuration.


TALISKER_CONFIG               A path to a python file containing configuration
                              variables.
DEVEL                         Allows coloured logs, warnings, and other
                              development convieniences.
TALISKER_COLOUR               Controls the coloured output of logs, can be 0
                              (off), 1 to enable, or 'simple' for a simpler
                              colourscheme.
DEBUGLOG                      Path to write debug level logs to, which is
                              enabled if path is writable.
TALISKER_SLOWQUERY_THRESHOLD  Set the threshold (in ms) over which SQL queries
                              will be logged. Defaults to -1 (off).
TALISKER_SOFT_REQUEST_TIMEOUT Set the threshold (in ms) over which WSGI requests
                              will log a warning-level sentry report. Defaults
                              to -1 (off).
TALISKER_LOGSTATUS            Sets whether http requests to /_status/ endpoints
                              are logged in the access log or not. Defaults to
                              false.
TALISKER_NETWORKS             Sets additional CIDR networks that are allowed to
                              access restricted /_status/ endpoints. Comma
                              separated list.
TALISKER_REVISION_ID          Sets the explicit revision of the application. If
                              not set, a best effort detection of VCS revision
                              is used.
TALISKER_UNIT                 Sets the instance name for use with sentry
                              reports.
TALISKER_ENV                  Sets the deployed environment for use with sentry
                              reports (e.g. production, staging).
TALISKER_DOMAIN               Sets the site domain name for use with sentry
                              reports.
STATSD_DSN                    Sets the Statsd DSN string, in the form:
                              udp://host:port/my.prefix
SENTRY_DSN                    Sets the sentry DSN, as per usual sentry client
                              configuration.
```

talisker/__init__.py
```diff
@@ -148,6 +149,69 @@ def run():
...
+    indent = 30
+    rest = width - indent
+    indent_str = '\n' + ' ' * indent
+    metadata = get_config().metadata()
+
+    if len(sys.argv) > 1:
+        name = sys.argv[1]
+        if name.upper() not in metadata:
+            sys.stderr.write('Invalid config: {}\n'.format(name))
+            sys.exit(1)
```
sparkiegeek: You can do this all in one with 
```suggestion
            sys.exit('Invalid config: {}'.format(name))
```

PR #390, opened 2019-01-07
fix parsing of start_time being a string in recent sentry

bloodearnest:


talisker/sentry.py
```diff
@@ -149,7 +149,14 @@ def add_talisker_context(data):
         user.pop('username', None)
 
     breadcrumbs = data.get('breadcrumbs', {}).get('values', [])
+
     start_time = data['extra'].get('start_time')
+    if start_time is not None:
+        try:
+            start_time = float(start_time)
```
sparkiegeek: I guess we're pretty confident we're not getting 'NaN' or 'Inf' from sentry?

bloodearnest: We add this ourselves  earlier. Sentry converts it to a string, as all 'extra' data is sent as strings. It used to do that on send, but recent version seems to do it eagerly on set.

PR #400, opened 2019-02-01
Pre move

bloodearnest:
This moves metrics and log metadata generation from gunicorn.py to wsgi.py. This is preparation for moving access logging from gunicorn to WSGI in general, but doesn't actually change any functionality, just where the code lives. The goal is to help reduce the diff size for the main PR.

### DIFF
--- a/talisker/wsgi.py
+++ b/talisker/wsgi.py
@@ -28,10 +28,12 @@ from __future__ import division
 from __future__ import absolute_import
 
 from builtins import *  # noqa
+__metaclass__ = type
+
 from collections import OrderedDict
 
 import talisker.request_id
-from talisker.context import CONTEXT
+import talisker.context
 import talisker.endpoints
 import talisker.statsd
 import talisker.requests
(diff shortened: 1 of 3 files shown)


talisker/wsgi.py
```diff
@@ -45,6 +48,30 @@
...
+    latency = talisker.metrics.Histogram(
+        name='wsgi_latency',
+        documentation='Duration of requests served by WSGI',
+        labelnames=['view', 'status', 'method'],
+        statsd='{name}.{view}.{method}.{status}',
+        buckets=[4, 8, 16, 32, 64, 128, 256, 512, 1024, 2048, 4096, 8192],
+    )
+
+    count = talisker.metrics.Counter(
+        name='wsgi_count',
```
tomwardill: Minor, but I kind of dislike this name. `wsgi_request_count` maybe?

tomwardill: Similar with the parameter name. `count` that is a `Counter` isn't very descriptive of what it's counting.

PR #402, opened 2019-02-01
Move access logging/metrics out of gunicorn and into WSGI middleware.

bloodearnest:
This makes us much less dependant on gunicorn, and provides Talisker's
enhanced logging for any WSGI server.

This also consolidates the separate middlewares we had for adding
environ items and headers into one main middleware, as well as adding
simple error handling at the WSGI app level, rather than letting the
WSGI handler errors, which gives us more control. The error response
here is very simple, and will improved in future work.

### DIFF
--- a/tests/test_wsgi.py
+++ b/tests/test_wsgi.py
@@ -29,38 +29,454 @@ from __future__ import absolute_import
 
 from builtins import *  # noqa
 
-from talisker.testing import run_wsgi
-from talisker import wsgi
+import time
+import wsgiref.util
+
+import pytest
+from freezegun import freeze_time
+
+from talisker import request_id, wsgi
+from talisker.context import track_request_metric
 
 
-def app(environ, start_response):
-    start_response(200, [])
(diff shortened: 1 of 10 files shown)


tests/test_wsgi.py
```diff
@@ -29,38 +29,454 @@
...
+    response.start_response('200 OK', [], None)
+
+    def iterator():
+        start_response.headers_sent = True
+        yield b'some content'
+        raise Exception('error')
+
+    it = response.wrap(iterator())
+    with pytest.raises(Exception):
+        list(it)
```
sparkiegeek: Am I missing something here? Where's the assert that `start_response.headers_sent` is `True`?

bloodearnest: headers_sent is part of the setup, not the assertion. By setting that, we alter the behaviour of mock_send_response to simulate the case when headers have been sent, in which case, the WSGI spec says the WSGI server should reraise the exc.

PR #405, opened 2019-02-12
Add requests.TaliskerAdapter, a requests transport adapter

bloodearnest:
It provides
  - load balancing with retries
  - default timeouts
  - timeout is for all retries, not per retry

It doesn't yet integrate witht the logging and metrics, and is not used
by default, but this is a first pass for further iteration.


talisker/requests.py
```diff
@@ -302,3 +310,137 @@ def enable_requests_logging():  # pragma: nocover
...
     requests_log.setLevel(logging.DEBUG)
     requests_log.propagate = True
+
+
+class TaliskerAdapter(HTTPAdapter):
+
+    def __init__(self, backends=None, connect=1.0, read=10.0, max_retries=0,
+                 *args, **kwargs):
+        # set up backends
+        if backends:
```
sparkiegeek: Extract this to a `_make_backend_iter()` or similar? It feels a bit too chunky for an `__init__`

talisker/requests.py
```diff
@@ -302,3 +310,137 @@ def enable_requests_logging():  # pragma: nocover
...
+        else:
+            self.__retry = max_retries
+
+        # disable lower level urllib3 retries completely
+        kwargs['max_retries'] = Retry(0, read=False)
+        super().__init__(*args, **kwargs)
+
+    def select_backend(self, request):
+        """Replaces the netloc of the url with a new netloc."""
+        next_endpoint = next(self.backend_iter)
```
sparkiegeek: If we've passed in an iterator, this could fall off the end and raise an ugly exception.

Think we should catch this and raise a nicer error? Or `itertools.cycle` the given iterator instead to ensure that we don't fall off the end?

test
```diff
@@ -0,0 +1 @@
+sdfa
```
sparkiegeek: errr?

tests/test_requests.py
```diff
@@ -25,10 +25,20 @@
...
-import pytest
+from collections import namedtuple
+import datetime
+import http.client
+import io
+import itertools
 import raven.context
-import requests
 import responses
+import requests
```
sparkiegeek: alphabetical sorting please `requests` < `responses`

tests/test_requests.py
```diff
@@ -319,3 +329,280 @@ def test_configured_session_with_user_name(context):
...
+        kws.update(kwargs)
+        return requests.Response()
+
+    monkeypatch.setattr(requests.adapters.HTTPAdapter, 'send', timeouts)
+    session.get('http://name/foo')
+    assert kws['timeout'] == (1.0, 10.0)
+
+
+class FakeSocket():
+    """Pretent to be read only socket-like object that implements makefile."""
```
sparkiegeek: s/Pretent/Pretend/ ?

PR #406, opened 2019-02-14
Move the logic from the sentry WSGI middleware into Talisker's middleware

bloodearnest:
This means:
  - cleaner implementation and less overhead
  - X-Sentry-ID header always included if there was a sentry report
  - we now support the wsgi.file_wrapper optimisation, which the sentry
  middleware breaks.


sparkiegeek:
:+1: but needs rebasing

PR #407, opened 2019-02-15
support rendering json responses as well as text and html

bloodearnest:


talisker/render.py
```diff
@@ -200,9 +256,29 @@ def format_table(self, table, widths):
                 w = widths[i]
                 if i < len(row) - 1:
                     w += 2
-                output.append('{:{w}}'.format(col, w=w))
+                output.append('{:<{w}}'.format(col, w=w))
```
bloodearnest: This was done to support widths of 0

talisker/render.py
```diff
@@ -226,3 +303,18 @@ def text_content(self):
...
+class PreformattedText(Content):
+    def __init__(self, *args, **kwargs):
+        kwargs.setdefault('tag', 'pre')
+        super().__init__(*args, **kwargs)
+
+    def text_content(self):
+        return self.content
+
+    def json_content(self):
+        return self.content.split('\n')
```
cjwatson: I'm not sure I understand the logic of this; I'd have thought that just leaving the content unsplit was a better JSON representation of preformatted text.  Is this just a test helper (in which case it should move to test code), or is there some rationale I'm missing?

bloodearnest: There is a rationale, although it might be questionable, and it isn't obvious as it's not used in this branch, but is in a follow on branch.

It makes it easier to read the preformatted content (e.g. traceback) at the cli when formatted with jq or similar, as a list is show once per line. And would still allow easy reconstruction if you want to display it. Bit of hack, I might drop it if you think the above is not worth it?

You're right that it's

PR #412, opened 2019-02-21
Track in flight requests and log any when gunicorn workers exit

bloodearnest:
This means we get a proper log message in the case of timeouts, and
removed the need for the previous pre_request hook to clear contexts

### DIFF
--- a/talisker/gunicorn.py
+++ b/talisker/gunicorn.py
@@ -31,6 +31,7 @@ from builtins import *  # noqa
 
 from collections import deque
 import logging
+import time
 
 from gunicorn.glogging import Logger
 from gunicorn.app.wsgiapp import WSGIApplication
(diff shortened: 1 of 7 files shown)


talisker/gunicorn.py
```diff
@@ -112,15 +113,16 @@ def gunicorn_child_exit(server, worker):
...
-    Clear any previous contexts on new request.
-
-    Note: we do this on way in, rather than the way out, to preserve the
-    request_id context when there's a timeout.
-    """
-    talisker.clear_contexts()
+def gunicorn_worker_exit(server, worker):
+    """Logs any requests that are still in flight."""
+    now = time.time()
+    for id, env in talisker.wsgi.REQUESTS.items():
```
sparkiegeek: drop `id` since it's not used, and iterate over `.values()`
```suggestion
    for env in talisker.wsgi.REQUESTS.values():
```

PR #420, opened 2019-03-05
fix werkzeug debugger and add test

bloodearnest:


talisker/wsgi.py
```diff
@@ -228,12 +228,13 @@ def start_response(self, status, headers, exc_info=None):
 
     def ensure_start_response(self, force=False):
         if force or not self.start_response_called:
-            self.original_start_response(
-                self.status,
-                self.headers,
-                self.exc_info,
-            )
-            self.start_response_called = True
+            if self.start_response_timestamp:
```
sparkiegeek: can't this just be added to an `and` on the line before?

bloodearnest: Yeah, but I think it reads clearer than a compound if. The logic it's embodying is a different question.

PR #428, opened 2019-03-12
support super lazy WSGI apps

bloodearnest:


tbille:
Tested on snapcraft.io codebase and works well for me.

PR #431, opened 2019-03-19
Fix no content responses by still trying to call start_response() early

bloodearnest:


talisker/wsgi.py
```diff
@@ -276,20 +275,33 @@ def __next__(self):
...
+        if self.status is not None and not self.start_response_called:
+            self.call_start_response()
+
         try:
             chunk = next(self.iter)
-            # some WSGI apps don't actually call start_response() until they've
-            # started iteration, so we delay calling the WSGI server's start
-            # response until the last possible moment
-            self.ensure_start_response()
+            # support lazy WSGI apps, as above
```
sparkiegeek: shouldn't this be in an `else:` clause - we don't expect `start_response()` to call `StopIteration` (or `GeneratorExit` or ... any of the others?)

bloodearnest: Hmm, calling the upstream start_response can raise, if the headers have already been sent, so my think it's better to keep it in the try: block, as that way we'd get a sentry report if it raises.

PR #437, opened 2019-04-02
add flaky to the test failing in travis

bloodearnest:


tests/test_gunicorn.py
```diff
@@ -184,6 +184,7 @@ def test_gunicorn_clears_context():
 
 
 @pytest.mark.timeout(120)
+@pytest.mark.flaky
```
sparkiegeek: can we somehow detect we're on travis and only mark it as flaky if we are?

PR #468, opened 2019-06-26
Add TALISKER_STATUS_INTERFACE config

### LINKED ISSUE #467: Alternative bind address for status endpoints
Would it be possible with gunicorn to have the `/_status` endpoints exposed over a different `interface:port` than the primary one?

Our use case is that we run an internal network where metrics are exposed separate from the primary one that user traffic is served off of.

bloodearnest:
This allows /_status endpoints to only respond on that interface.

Includes refactor of GunicornProcess to support binding to multiple
interfaces.

Fixes #467

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -48,7 +48,7 @@ lint: $(VENV)
 	$(BIN)/flake8 talisker tests
 
 _test: $(VENV)
-	. $(BIN)/activate && $(BIN)/pytest -n auto --timeout=15 $(ARGS)
+	. $(BIN)/activate && $(BIN)/pytest -n auto --timeout=15 --no-success-flaky-report $(ARGS)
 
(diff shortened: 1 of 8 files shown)


requirements.tests.txt
```diff
@@ -3,6 +3,7 @@ pytest==4.4.1
 freezegun==0.3.11
 pytest-cov==2.6.1
 pytest-postgresql==1.4.0
+mirakuru==1.1
```
maxiberta: This seems to fix a crash when attempting to run tests.

`TypeError: descriptor '__subclasses__' of 'type' object needs an argument`

bloodearnest: Yeah, it was an unpinned dependency of pytest-postgresql, and they just release 2.0, which breaks in py2, among other issues.

PR #492, opened 2019-12-04
refactor wsgi request logging, for more context

bloodearnest:
Renames WSGIResponse to TaliskerWSGIRequest, as it's not really just a response. Moves the logic of request logging/metrics into the class, so more context is available. Later changes will use this extra context to enhance sentry reports.

### DIFF
--- a/talisker/testing.py
+++ b/talisker/testing.py
@@ -283,11 +283,11 @@ class TestContext():
         self.sentry_context = talisker.sentry.TestSentryContext(self.dsn)
 
     def start(self):
-        Context.new()
         self.old_statsd = talisker.statsd.get_client.raw_update(
             self.statsd_client)
         talisker.logs.add_talisker_handler(logging.NOTSET, self.handler)
         self.sentry_context.start()
+        Context.new()
 
     def stop(self):
         logging.getLogger().handlers.remove(self.handler)
--- a/talisker/wsgi.py
+++ b/talisker/wsgi.py
@@ -66,7 +66,10 @@ REQUESTS = {}
 
 
 def talisker_error_response(environ, headers, exc_info):
-    """Returns WSGI iterable to be returned as an error response
+    """Returns WSGI iterable to be returned as an error response.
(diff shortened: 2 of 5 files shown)


talisker/testing.py
```diff
@@ -283,11 +283,11 @@ def __init__(self, name=None):
         self.sentry_context = talisker.sentry.TestSentryContext(self.dsn)
 
     def start(self):
-        Context.new()
         self.old_statsd = talisker.statsd.get_client.raw_update(
             self.statsd_client)
         talisker.logs.add_talisker_handler(logging.NOTSET, self.handler)
         self.sentry_context.start()
+        Context.new()
```
bloodearnest: FTR - this is a driveby UX improvement - the previous lines generates logs, moving the new context to here means those logs are not captured as part of the context.

talisker/wsgi.py
```diff
@@ -398,16 +390,113 @@ def finish_request(self):
...
+            WSGIMetric.latency.observe(extra['duration_ms'], **labels)
+            if timeout:
+                lbls = labels.copy()
+                lbls.pop('status')
+                WSGIMetric.timeouts.inc(**lbls)
+            elif self.status_code and self.status_code >= 500:
+                WSGIMetric.errors.inc(**labels)
+
+        return extra
+
```
bloodearnest: These two methods are virtually unchanged from their module function versions.

talisker/wsgi.py
```diff
@@ -457,135 +546,32 @@ def __call__(self, environ, start_response):
...
 
-        response = WSGIResponse(environ, start_response, self.headers)
+        # track in-flight requests
+        if rid in REQUESTS:
+            logger.warning(
+                'duplicate request id received by gunicorn worker',
+                extra={'request_id': rid}
+            )
+        else:
+            REQUESTS[rid] = request
```
bloodearnest: We now track the entire request object, rather than just the environ

PR #493, opened 2019-12-11
Sentry logging breadcrumb fixes

bloodearnest:
Ensure that there is an active sentry context by activating it before
each request. Previously, there was no active context for the very first
request a worker served.

Also improves logging breadcrumbs:
 - remove duplicated information
 - condense file/lineno/funcName into one item
 - stringify keys so breadcrumb displays properly

Cleans up the depreacted use of LogRecord._structured

### DIFF
--- a/talisker/sentry.py
+++ b/talisker/sentry.py
@@ -54,6 +54,7 @@ __all__ = [
     'get_client',
     'record_breadcrumb',
     'report_wsgi',
+    'new_context',
     'set_client',
 ]
 
@@ -75,6 +76,9 @@ try:
     import raven.handlers.logging
 except ImportError:
     # dummy APIs that do nothing
+    def new_context():
+        pass
+
     def record_breadcrumb(*args, **kwargs):
         pass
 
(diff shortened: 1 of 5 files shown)


talisker/sentry.py
```diff
@@ -141,14 +153,18 @@ def record_log_breadcrumb(record):
...
-                'path': record.pathname,
-                'lineno': record.lineno,
+                'location': '{}:{}'.format(record.pathname, record.lineno)
             }
-            if hasattr(record, 'func'):
-                metadata['func'] = record.func
-            metadata.update(getattr(record, '_structured', {}))
+            funcName = getattr(record, 'funcName', None)
+            if funcName:
+                metadata['location'] += ':' + record.funcName + '()'
```
tonysimpson: I feel the file:line:func format is pretty common and recognisable and so the '()' postfix is not needed (I haven't seen that syntax used before).
Also perhaps there is an argument for having the second colon even if you can't get the function name as it might make it easier to parse by eye or in scripts.

bloodearnest: Good call, will do

PR #494, opened 2019-12-12
Fix thread local not being reset on fork

bloodearnest:
This particularly affected the requests session.

### DIFF
--- a/talisker/requests.py
+++ b/talisker/requests.py
@@ -35,7 +35,6 @@ import functools
 import itertools
 import logging
 import random
-import threading
 import warnings
 import time
 from future.moves.urllib.parse import (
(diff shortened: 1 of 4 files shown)


bloodearnest:
FTR, I actually tried doing a os.fork() in the tests, which worked, but would hang when run under pytest-xdist. So I gave up and simulated it. But it does work IRL too, just couldn't find a way to get that part of the tests.

bloodearnest:
> FTR, I actually tried doing a os.fork() in the tests, which worked, but would hang when run under pytest-xdist. So I gave up and simulated it. But it does work IRL too, just couldn't find a way to get that part of the tests.

I figured out the issue, and have updated the tests to use os.fork() properly.

PR #499, opened 2020-01-07
Add EXPLAIN output to sql sentry breadcrumbs

bloodearnest:
Also fixes a bug where the name of the duration field was wrong in tests.

### DIFF
--- a/talisker/config.py
+++ b/talisker/config.py
@@ -129,6 +129,7 @@ class Config():
         'TALISKER_NETWORKS': [],
         'TALISKER_ID_HEADER': 'X-Request-Id',
         'TALISKER_DEADLINE_HEADER': 'X-Request-Deadline',
+        'TALISKER_EXPLAIN_SQL': False,
     }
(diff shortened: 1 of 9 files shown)


wgrant:
I don't think this should be done unconditionally, or by default. It's quite expensive, as you end up doing twice as many queries as you otherwise would.

bloodearnest:
I was just writing a comment to this effect,

To clarify, the extra queries are a) only done after the request has finished b) only if a sentry report is generated, so infrequently. The main time this would be an issue AIUI is adding additional load to the DB when there is an outage that is generating lots of sentry requests, which is already a problem. 

What condition do you think might work?

We are adding the X-Debug header in #498, perhaps we only do EXPLAIN when that is requested?

We could also maybe cache the EXPLAIN results per parameterised query string in LRU cache. Given it's not ANALYSE, the plan returned should be the same for all queries. And query structured should be fairly stable, per process launch.

PR #502, opened 2020-02-12
support gunicorn 20.x

bloodearnest:


nottrobin:
Thanks for the quick work @bloodearnest!

PR #517, opened 2020-04-17
fix missing rogue import

bloodearnest:


talisker/testing.py
```diff
@@ -51,6 +51,7 @@
 import talisker.requests
 import talisker.statsd
 import talisker.util
+import talisker.sentry
```
sparkiegeek, line 54: sort the imports!

bloodearnest: :stuck_out_tongue:

PR #543, opened 2020-09-03
Require contexts to be explicitly created.

bloodearnest:
Implicitly creating contexts causes a variety of issues, and was a bit
of a hack when first doing the contextvars work.

This change introduces a NullContext, which is the default unless
you've explicitly created one. This NullContext provides the same API as
a context, but does not store any information.

This fixes a few different issues, from context bleeding and leaking,
through to test fixture issues that clear the context dict.

This is an internal API change, and shouldn't affect users code.

### DIFF
--- a/talisker/context.py
+++ b/talisker/context.py
@@ -32,9 +32,9 @@ __metaclass__ = type
 
 import future.utils
 try:
-    from collections.abc import Mapping
+    from collections.abc import Mapping, Sequence
 except ImportError:  # py2
-    from collections import Mapping
+    from collections import Mapping, Sequence
 
 from collections import OrderedDict, defaultdict
 from contextlib import contextmanager
@@ -50,7 +50,10 @@ from talisker.util import early_log, pkg_is_installed
 __all__ = ['Context']
 
 
-CONTEXT_MAP = {}  # global storage for contexts by id
+# Global storage for contexts by id. We use a process global, so that we can
+# provide best effort logging of outstanding requests when a process is killed,
+# e.g. worker killed by master whilst having inflight requests
+CONTEXT_MAP = {}
 
 
 if future.utils.PY3:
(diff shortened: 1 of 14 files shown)


talisker/context.py
```diff
@@ -204,19 +331,22 @@ def delete_context(context_id):
...
 
 
+def warn_null_context(api, extra):
+    import logging
+    logging.getLogger('talisker.context').warning(
+        "{} API called when there is no active context, "
+        "data will not be stored".format(api),
+        extra=extra,
+    )
+    pass
```
sparkiegeek, line 341: unnecessary pass

talisker/context.py
```diff
@@ -234,127 +364,80 @@ def new(self):
         self.clear()
         ctx = create_context()
         ContextId.set(ctx.id)
+        print('new context: {}'.format(ctx.id))
```
sparkiegeek, line 367: KIWF

talisker/context.py
bloodearnest: Yep, it's a draft WIP :)

talisker/context.py
```diff
@@ -321,37 +368,71 @@ def new(self):
...
+    def soft_timeout(self):
+        return self.current().soft_timeout
+
+    @soft_timeout.setter
+    def soft_timeout(self, timeout):
+        current = self.current()
+        if current is not NULL_CONTEXT:
+            current.soft_timeout = timeout
+
+    def set_relative_deadline(self, timeout):
```
tonysimpson, line 402: nice!

tonysimpson (commented):
I wonder why we are exposing current? maybe _current would be better if we would rather users use the ContextAPI methods?

bloodearnest:
> I wonder why we are exposing current? maybe _current would be better if we would rather users use the ContextAPI methods?

So, it is used by other parts of talisker, as it needs to access the context data directly.

I'm not entirely happy with this API atm, tbh, needs a rethink. But I want to fix this bug.

Issue #556, opened 2020-12-08
support newer celery (>= 4.4)


verterok:
Hi, would be great to support a newer version of celery, I'm looking at 4.4.2 for "Task class definitions can have retry attributes" in particular.

Will try to get a PR ready in the next couple of days.

Cheers

woutervb:
see #574 for an update

PR #557, opened 2020-12-08
support celery<=4.4.3

verterok:
allow/support celery<4.4.3

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -53,7 +53,7 @@ install_requires =
 gunicorn = gunicorn>=19.7.0,<21.0
 raven = raven>=6.4.0,<7.0
 celery = 
-	celery>=3.1.25.0,<4.4
+	celery>=3.1.25.0,<4.4.3
 	kombu>=3.0.37,<4.5
 django = django>=1.10,<2.0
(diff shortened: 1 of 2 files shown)


verterok:
fixes #556

maxiberta (commented):
Thanks! `<4.4.3` seems to contradict the PR title. Still, why not just `<5.0` so that it's compatible with eg. celery 4.4.7?

verterok:
> Thanks! `<4.4.3` seems to contradict the PR title. Still, why not just `<5.0` so that it's compatible with eg. celery 4.4.7?

fixed the title, I tested IRL with 4.4.2 :)
can't assure things work with a different version

Roadmaster (approved):
LGTM the title and code are now consistent and it seems OK not to go overboard with supported versions if only 4.4.2 and lower was tested.

maxiberta (approved):
Fair enough +1

bloodearnest:
Yo.

You may want to bump the minimum supported kombu/celery versions, there was some incompatibility with older versions that prevented supporting newer versions of kombu IIRC. We had to maintain support for the older version because SCA used it I think.

PR #560, opened 2021-05-21
Celery 5 support

hackoder:


talisker/celery.py
```diff
@@ -183,9 +183,30 @@ def task_prerun(sender, task_id, task, **kwargs):
...
+        # the call stack ends up looking something like:
+        # pre -> pre -> pre -> post -> post -> post
+
+        # In celery 5.x, the behavior has been fixed and works similarly
+        # in both eager and non-eager mode
+
+        # To emulate non-eager behavior in eager mode for celery 4.x, we check
+        # if the request is eager (task.request.is_eager) and its a retry (
+        # task.talisker_timestamp is already set). If so, we manually track the
+        # calls to retry and task-post here.
```
bloodearnest, line 207: Hoo, boy, that reads like a lot of fun to debug.

I hope celery 5.0 has cleaned up some of the magic in its internals, because it was crazy.

hackoder: Heh, you're right about the fun part!

5.0 is much better, and behavior is more consistent in eager/non-eager mode. I noticed in this specific bit of code that the retry and latency metrics were being recorded in the reverse order. Just flipping them around allows tests to pass in 3.x, 4.x and 5.x, and the behavior is the same in both eager/async mode.

talisker/celery.py
```diff
@@ -183,9 +183,30 @@ def task_prerun(sender, task_id, task, **kwargs):
...
+        # In celery 4.x:
+        # When the task is called eagerly, because the call is synchronous
+        # the call stack ends up looking something like:
+        # pre -> pre -> pre -> post -> post -> post
+
+        # In celery 5.x, the behavior has been fixed and works similarly
+        # in both eager and non-eager mode
+
+        # To emulate non-eager behavior in eager mode for celery 4.x, we check
+        # if the request is eager (task.request.is_eager) and its a retry (
```
maxiberta, line 205: `s/its/it's/`. Also, check the trailing parens.

maxiberta (approved):
LGTM!

PR #562, opened 2021-06-08
Add GH Actions support

hackoder:
Note that the current travis configuration is to run slightly different tox environments depending on python version (lint is run against py27 and py35, docs against py35 etc). This PR aims to replicate the current behavior in GH actions, and uses [tox-gh-actions](https://github.com/ymyzk/tox-gh-actions) to achieve this.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -112,7 +112,7 @@ travis: $(VENV_PATH)
 	$(MAKE) $(LIMBO_REQUIREMENTS)
 	env/bin/tox
 
-gh-tox:
+github-tox:
 	pip install tox setuptools $(subst requirements,-c requirements,$(REQUIREMENTS))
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -112,6 +112,11 @@ travis: $(VENV_PATH)
 	$(MAKE) $(LIMBO_REQUIREMENTS)
 	env/bin/tox
 
+gh-tox:
```
Roadmaster, line 115: Since this will only ever be used from github, and not typed manually, I would suggest not abbreviating it to make it clearer (github-tox).

Roadmaster (approved):
Looks OK! Made one optional suggestion but we can also merge as-is and iterate.

Makefile
hackoder: ack, good idea!

PR #565, opened 2021-06-25
Send view-name (if available) to sentry with soft timeout messages

suligap:
This should make the default grouping of those events in Sentry more
accurate in absence of other attributes like stacktrace in this event.
Particularly for views with url path args.

This changes both the "message" itself and adds a transaction tag/field
to the event if it's not set (it might not be available in this context
for example with Django when the appropriate middleware had already
cleared the transaction name).

### DIFF
--- a/.github/workflows/tox.yml
+++ b/.github/workflows/tox.yml
@@ -34,6 +34,7 @@ jobs:
           python-version: ${{ matrix.python }}
       - name: Install dependencies
         run: |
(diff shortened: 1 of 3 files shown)


bloodearnest:
Nice idea!

maxiberta (approved):
LGTM! +1

suligap:
Thanks all!

PR #568, opened 2021-09-04
Make util.Local threadsafe and extend its tests

suligap:
This threading.local wrapper was "forksafe" but didn't have the threading.local's capabilities anymore, due to a bug which wasn't obvious if only hasattr() was used to access its attrs (like in the tests here).

### DIFF
--- a/talisker/util.py
+++ b/talisker/util.py
@@ -276,7 +276,11 @@ class Local(object):
 
     def _check(self):
         pid = os.getpid()
-        if self._local._pid != pid:
+        if not hasattr(self._local, '_pid'):
+            # In a new thread - just add the missing _pid attr.
+            self._local._pid = pid
+        elif self._local._pid != pid:
+            # In a new process - discard the thread local.
             self._local = threading.local()
             self._local._pid = pid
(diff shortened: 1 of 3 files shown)


woutervb (approved):
Thanks for the update, this does indeed explain, and fix some of the observed unexpected behaviour

bloodearnest:
Oh dear. 🤦‍♂️ 

Did you see some strange behaviour due to this?

suligap:
> Did you see some strange behaviour due to this?

Failures in tests under testtools' concurrent test suite. Fortunately, it was raising an AttributeErrror once trying to access an attribute with anything other than hasattr(). Threads aren't very popular so it went undetected for some time :)

bloodearnest:
Interesting. 

Previously, talisker did use threads, as it used raven's default transport, which was uses a background thread to send the sentry report. In #496, it switched to a deferred synchronous transport, so I suspect there hasn't been much production threads usage since then.

suligap:
That makes sense. Both the custom Local and the sentry transport change are in the same release.

PR #569, opened 2021-09-06
Allow django <4 as it is known that this works

woutervb:


woutervb:
Re-open as it will work once python 2.7 support is dropped

maxiberta (commented):
I'm curious what changed in talisker to enable support for django<4.0? And why bump the min version?

maxiberta:
> I'm curious what changed in talisker to enable support for django<4.0? And why bump the min version?

Oh I see, dropping support for python 2.7 in #570. I would merge both commits in a single branch/PR to ensure that both land together.

PR #570, opened 2021-09-06
Remove support for python 2.7

woutervb:


setup.cfg
```diff
@@ -29,10 +29,12 @@ classifiers =
 	Topic :: Internet :: WWW/HTTP :: WSGI
 	Topic :: Internet :: WWW/HTTP :: WSGI :: Middleware
 	Topic :: System :: Logging
-	Programming Language :: Python :: 2.7
 	Programming Language :: Python :: 3.4
```
maxiberta, line 32: I think 3.4 shoud be dropped as well?

setup.py
```diff
@@ -116,10 +116,12 @@
         'Topic :: Internet :: WWW/HTTP :: WSGI',
         'Topic :: Internet :: WWW/HTTP :: WSGI :: Middleware',
         'Topic :: System :: Logging',
-        'Programming Language :: Python :: 2.7',
         'Programming Language :: Python :: 3.4',
```
maxiberta, line 119: I think 3.4 shoud be dropped as well?

maxiberta (approved):
+1 with a little comment about dropping python 3.4

PR #571, opened 2021-09-06
Ensure that newer versions of celery can be used

woutervb:
No need to need to force older versions of celery when python < 3.6


maxiberta (approved):
LGTM

PR #574, opened 2021-09-21
Remove python 2.7 add python 3.7, 3.8, 3.9, 3.10

woutervb:
Removal of python 2.7 references

Added python 3.7, 3.8, 3.9 and 3.10

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -59,6 +59,10 @@ a common operational platform for your python microservices.
 It integrates with many standard python libraries to give you out-of-the-box
 logging, metrics, error reporting, status urls and more.
 
+Python version support
+----------------------
+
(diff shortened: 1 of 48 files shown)


setup.py
```diff
@@ -59,6 +59,10 @@
 It integrates with many standard python libraries to give you out-of-the-box
 logging, metrics, error reporting, status urls and more.
 
+Python version support
+----------------------
+
+This release of talisker (0.20.0) will be the last to support python 2.7
```
maxiberta, line 65: Oh I see, this looked a bit confusing at first but I see the README file contains this warning already.

maxiberta (approved):
LGTM, thanks! However, should we update the GH CI workflow as well? Approving anyway, let CI have the last word :-)

PR #579, opened 2021-11-03
In LogRecord the extra field is optional

woutervb:
LogRecord fields don't always have the extra field set, to always
referring to is causes Attribute exceptions

### DIFF
--- a/requirements.tests.txt
+++ b/requirements.tests.txt
@@ -19,7 +19,8 @@ eventlet==0.30.2  # pyup: <0.30.3
 gevent==20.9.0
 # the bottom pin is for limbo test runs, as latest version doesn't work with
 # newer celery versions
-redis>=2.10.6,<=3.2.1
+redis>=3.4.1;python_version > "3.5"
+redis>=2.10.6,<=3.2.1;python_version <= "3.5"
 SQLAlchemy==1.3.13
 # newer versions of celery pin vine versions, however
 # older versions don't. This pin is for limbo test runs for
--- a/requirements.tests.txt
+++ b/requirements.tests.txt
@@ -19,8 +19,7 @@ eventlet==0.30.2  # pyup: <0.30.3
 gevent==20.9.0
 # the bottom pin is for limbo test runs, as latest version doesn't work with
 # newer celery versions
-redis>=3.4.1;python_version > "3.5"
-redis>=2.10.6,<=3.2.1;python_version <= "3.5"
+redis>=2.10.6
 SQLAlchemy==1.3.13
 # newer versions of celery pin vine versions, however
 # older versions don't. This pin is for limbo test runs for
--- a/setup.py
+++ b/setup.py
@@ -147,6 +147,7 @@ setup(
         celery=[
             'celery>=3.1.25.0,<4.4.7;python_version<="3.5"',
             'celery>=4,<5.3;python_version>"3.5"',
+            'kombu<5.2.0',
         ],
         dev=[
             'logging_tree>=1.7',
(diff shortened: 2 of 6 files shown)


setup.py
```diff
@@ -147,6 +147,7 @@
         celery=[
             'celery>=3.1.25.0,<4.4.7;python_version<="3.5"',
             'celery>=4,<5.3;python_version>"3.5"',
+            'kombu<5.2.0',
```
suligap, line 150: It looks like kombu<6 are compatible with celery 5.1.2: https://github.com/celery/celery/blob/v5.1.2/requirements/default.txt#L3

And it also looks like the errors were coming from the `redis` client. Did you try bumping the `redis` package version in requirements.tests.txt instead?

woutervb: Determined the minimum version of redis that supports the username parameter and added that to the `requirements.test.txt` file

requirements.tests.txt
```diff
@@ -19,7 +19,8 @@ eventlet==0.30.2  # pyup: <0.30.3
 gevent==20.9.0
 # the bottom pin is for limbo test runs, as latest version doesn't work with
 # newer celery versions
-redis>=2.10.6,<=3.2.1
+redis>=3.4.1;python_version > "3.5"
+redis>=2.10.6,<=3.2.1;python_version <= "3.5"
```
suligap, line 23: It looks like we could be good with replacing both of the above with:
```
redis>=2.10.6
```
py35 passes with the latest `redis` version.

And the comment above could be updated as well I suppose?

talisker/testing.py
```diff
@@ -131,7 +131,7 @@ def cmp(a, b):
                 return False
 
         if all(cmp(getattr(record, k, _s), v) for k, v in kwargs.items()):
-            if extra:
+            if extra and hasattr(record, 'extra'):
```
suligap, line 134: It'd be nice to add a regression test for this. Like

```python
def test_log_record_list_no_extra():
    makeRecord = logging.Logger('test').makeRecord

    def record(name, level, msg, extra=None):
        return makeRecord(name, level, 'fn', 123, msg, None, None)

    r1 = record('root.log1', logging.WARNING, 'foo')
    r2 = record('root.log2', logging.INFO, 'foo bar')
    records = testing.LogRecordList()
    records.extend([r1, r2])

    assert records.filter(name='root.log1') == [r1]
    assert records.filter(msg='foo') == [r1, r2]
    assert records.filter(level=logging.WARNING) == [r1]
    assert records.filter(extra={'a': 1}) == []
```

in tests/test_testing.py.

And this test fails the last assertion because the fix above results in `filter(extra=...)` matching all records that do not have `.extra`. So I think we should return False here if `extra` is passed in `filter()` but the record doesn't have `.extra`. What do you think?

suligap (changes requested):

requirements.tests.txt
woutervb: Yes, but the test for minimal version won`t as there an older version of kombu is allowed, so this change is not an options without bumping minimal requirements

tests/test_logging.py
```diff
@@ -142,6 +143,23 @@ def test_make_record_all_extra():
     assert record._structured == {'a': 1, 'b': 2, 'c': 3}
 
 
+def test_log_record_list_no_extra():
```
suligap, line 146: This test should be moved to test_testing.py.

suligap (approved):
Thanks for the updates. Approving conditionally on the test case move to test_testing.py.

PR #586, opened 2022-08-25
Fix the test prometheus counter

woutervb:
When referring to the test_counter, removing the reference to the
registry ensures that it works both with single or multiple workers.

### DIFF
--- a/talisker/endpoints.py
+++ b/talisker/endpoints.py
@@ -108,8 +108,8 @@ def test_counter():
     # Also, isolate in its own registry. Multiprocess mode doesn't really use
     # registries, but it helps keep things separated, which is useful in
     # testing.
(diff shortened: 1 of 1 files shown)


talisker/endpoints.py
```diff
@@ -108,8 +108,8 @@ def test_counter():
     # Also, isolate in its own registry. Multiprocess mode doesn't really use
     # registries, but it helps keep things separated, which is useful in
```
maxiberta, line 109: I think this comment is obsolete and can be dropped now?

maxiberta: Anyway, tested locally and it works. Thanks!

maxiberta (approved):
LGTM (maybe drop/update that comment)

PR #587, opened 2022-08-26
Reduce version complexity

woutervb:
Reduce supported python versions to those that come with python lts releases

* python3.5 (xenial)
* python3.6 (bionic)
* python3.8 (focal)

### DIFF
--- a/.github/workflows/tox.yml
+++ b/.github/workflows/tox.yml
@@ -24,7 +24,7 @@ jobs:
     runs-on: ubuntu-20.04
     strategy:
       matrix:
-        python: [2.7, 3.5, 3.6, 3.7, 3.8, 3.9]
+        python: [3.5, 3.6, 3.8]
 
     steps:
       - uses: actions/checkout@v2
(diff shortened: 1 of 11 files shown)


sparkiegeek:
Ubuntu 22.04 LTS comes with Python 3.10 - it shouldn't be excluded?

.github/workflows/tox.yml
```diff
@@ -24,7 +24,7 @@ jobs:
     runs-on: ubuntu-20.04
     strategy:
       matrix:
-        python: [2.7, 3.5, 3.6, 3.7, 3.8, 3.9]
+        python: [3.5, 3.6, 3.8]
```
sparkiegeek, line 27: this is at odds with the idea of supporting the Python versions that come with Ubuntu LTS releases (since we always run on 20.04 with differing versions)

woutervb: Well it doesn't work, never had, so nothing is dropped here. It is fixed with a follow-up pr

maxiberta (approved):
+1 provided with merge together with the follow-up PR (#588) adding support for python 3.10.

PR #591, opened 2023-01-06
Fix spelling of Python 3.5-specific requirements

cjwatson:
According to PEP 440, `~="3.5"` is equivalent to `>="3.5", =="3.*"` - that is, any 3.x version from 3.5 on.  I don't think that's what was intended in most of these cases, and the effect is that resolvers on recent versions of Python end up pinning to the versions that work on 3.5, which is pretty excessive.

In most of these cases, `~="3.5.0"` is a better spelling, restricting these pins to Python 3.5.x.  The exceptions seem to be `aiocontextvars` and `contextvars`, which were needed in Python 3.5 and 3.6, but aren't needed as of 3.7 since `contextvars` is in the standard library there; I think `>="3.5"` plus `<"3.7"` and similar makes more sense for those cases. (Comments in `talisker.context` indicate that `aiocontextvars` specifically requires at least 3.5.3.)

### DIFF
--- a/.github/workflows/tox.yml
+++ b/.github/workflows/tox.yml
@@ -37,7 +37,7 @@ jobs:
           sudo apt update
           cat dependencies.txt | xargs sudo apt install -y
           python -m pip install --upgrade pip
-          pip install tox tox-gh-actions
+          pip install 'tox<4' 'tox-gh-actions<3'
       - name: Start redis
         run: sudo systemctl start redis-server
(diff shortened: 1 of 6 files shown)


tartley (approved):
lgtm!

verterok (approved):
looks good

maxiberta (approved):
LGTM!

PR #592, opened 2023-01-10
Drop the future library

cjwatson:
We no longer need it having dropped support for Python 2.


maxiberta (approved):
This looks great. Thanks for the cleanup! +1

PR #605, opened 2024-07-17
Fix building docs

jugmac00:

PR #607, opened 2024-09-02
update dependencies and add support for Python 3.12

### LINKED ISSUE #606: Adding support for Python 3.12
The latest LTS version of Ubuntu Noble 24.04 is shipped with Python3 version 3.12.
This package doesn't support 3.12 as importing modules like `talisker.flask` throws a not found module error.

Are there any plans to support this Python3 version?

goulinkh:
Fixes #606

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -118,6 +118,7 @@ setup(
         'Programming Language :: Python :: 3.6',
         'Programming Language :: Python :: 3.8',
         'Programming Language :: Python :: 3.10',
+        'Programming Language :: Python :: 3.12',
         'Programming Language :: Python :: Implementation :: CPython',
     ],
     description='A common WSGI stack',
@@ -138,7 +139,8 @@ setup(
         ],
         celery=[
             'celery~=4.4;python_version~="3.5.0"',
-            'celery>=4,<5.3;python_version>"3.5"',
+            'celery>=4,<5.3;python_version>"3.5" and python_version<"3.12"',
+            'celery>=5,<6;python_version>="3.12"',
         ],
         dev=[
             'logging_tree>=1.9',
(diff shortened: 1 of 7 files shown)


jugmac00 (approved):
Thanks for looking into this!

The failing docs job is unrelated, see https://github.com/canonical-ols/talisker/pull/605

setup.py
```diff
@@ -152,7 +154,8 @@
         ],
         flask=[
             'flask~=1.1;python_version~="3.5.0"',
-            'flask<3;python_version>"3.5"',
+            'flask<3;python_version>"3.5" and python_version<"3.12"',
+            'flask>3,<4;python_version>="3.12"',
```
maxiberta, line 158: What prevents Flask 2.x from working on python 3.12? Same question applies to all the other updated deps.

maxiberta (commented):
Just left a late question about pinned deps. Thx

setup.py
jugmac00: @goulinkh Good question which I have not thought about. Any reasons? We can anytime remove the strict restriction if things work with older versions, too.

goulinkh: Sorry for the delayed response.

I think I misunderstood the version requirement role, If understand correctly this should enforce a recent version but instead have a wide range of available version with the only limitation the minimal working python version.
I've done some research and have done a new version of this, let me know if this new version is valid for `Flask` and `Werkzeug` so I can continue with this logic on the rest of the changed requirements:
```
	Werkzeug~=1.0;python_version~="3.5.0"
	Werkzeug<2.1;python_version>="3.6" and python_version<"3.7"
	Werkzeug<2.3;python_version>="3.7" and python_version<"3.8"
	Werkzeug<3.1;python_version>="3.8"
	
        flask~=1.1;python_version~="3.5.0"
	flask<2.1;python_version>=3.6 and python_version<"3.7"
	flask<2.3;python_version>=3.7 and python_version<"3.8"
	flask<3.1;python_version>=3.8
```

goulinkh: This has been addressed here: https://github.com/canonical-ols/talisker/pull/608

PR #608, opened 2024-11-04
Remove unnecessary package versions constraints

goulinkh:


setup.cfg
```diff
@@ -98,4 +93,4 @@ console_scripts =
 	talisker.gunicorn.eventlet=talisker:run_gunicorn_eventlet[gunicorn]
 	talisker.gunicorn.gevent=talisker:run_gunicorn_gevent[gunicorn]
 	talisker.celery=talisker:run_celery[celery]
-	talisker.help=talisker:run_help
+	talisker.help=talisker:run_help
```
jugmac00, line 96: This seems to be an unwanted change. You probably deleted at the empty line at the end of the file.

setup.py
```diff
@@ -211,4 +206,4 @@
     url='https://github.com/canonical-ols/talisker',
     version='0.21.3',
     zip_safe=False,
-)
+)
```
jugmac00, line 209: This seems to be an unwanted change. You probably deleted at the empty line at the end of the file.

jugmac00 (approved):
This looks like a good first step - for me it seems it removes the recently added version constraints and mostly goes back to what has been in place before.

In the near future we should evaluate how to make the package compatible with Flask 3 / Werkzeug 3, but for now this looks good.

@maxiberta Could you please review the MP?

maxiberta (approved):
Looks good to me. Thanks for reverting such stringent requirements.

PR #610, opened 2024-12-04
Prevent prometheus_client deadlocks during signal handling by the gunicorn arbiter

suligap:
We have been affected by deadlocks caused by talisker instrumenting the gunicorn main process' logging with sentry and related prometheus metrics. The deadlocks were on a lock in prometheus_client.

So this is a single threaded context in which the main process was sometimes in the middle of incrementing a prometheus counter as a result of sending an event to sentry as a result of gunicorn main process logging an error log about eg. terminating an unresponsive worker. In the middle of that when the global multiprocessing mode prometheus client lock in values.py was held, a signal handler for SIGCHLD was invoked in that main process in response to some other worker terminating. During that signal handling the main process also logs an error message which caused the sentry event and a corresponding prometheus counter update -> deadlock.

So in order to be more careful/conservative with what we do during signal handling, or in this case what we instrument gunicorn to do, this change sets up all of the requests related metrics to not be emitted from the process they were created in (which for these metrics is the gunicorn arbiter if we're running gunicorn).

More details on prometheus client behavior in
https://github.com/prometheus/client_python/pull/1076

### DIFF
--- a/talisker/metrics.py
+++ b/talisker/metrics.py
@@ -25,6 +25,7 @@
 from contextlib import contextmanager
 import functools
 import logging
+import os
 import time
 
 import talisker.statsd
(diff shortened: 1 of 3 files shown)


suligap:
Closing in favor of #611

PR #614, opened 2025-03-14
Bump supported flask, werkzeug and statsd versions

edlerd:


maxiberta (commented):
Thanks for your contribution!
Would you please elaborate on why these new constraints are `python_version>=3.9`?

requirements.docs.txt
```diff
@@ -1,2 +1,3 @@
 Sphinx==5.3.0;python_version=="3.6"
 Sphinx==7.0.1;python_version>"3.6"
+werkzeug==2.3.8
```
maxiberta, line 3: Why this specific version of Werkzeug for docs requirements?

edlerd: This one was failing before because of a missing werkzeug. You are right, we don't have to specify a version for it here though. Relaxed this to just list werkzeug here.

maxiberta:
> > Thanks for your contribution! Would you please elaborate on why these new constraints are `python_version>=3.9`?
> 
> This is due to what the pages on https://pypi.org/project/Werkzeug/ and https://pypi.org/project/Flask/ list. For recent version, both say "Requires: Python >=3.9".

That applies to the latest release only. But for example, according to [Flask release notes](https://flask.palletsprojects.com/en/stable/changes/#version-3-1-0/) Python 3.8 was supported up to version 3.1.0. As a library, we should try to minimize restrictions really. In this case, I'd say just `;python_version>"3.8"`.

edlerd:
Went back to requiring python 3.9 as the dependencies could not be resolved with the relaxed requirements. [Python 3.8 is end of life](https://devguide.python.org/versions/), so we probably don't need to cater for it anyways.

maxiberta:
I'm attempting to env/bin/pip install -e . and got:
```
ERROR: Cannot install talisker==0.21.5 because these package versions have conflicting dependencies.

The conflict is caused by:
    talisker 0.21.5 depends on Werkzeug<3
    talisker 0.21.5 depends on Werkzeug>=3<4
```

Not sure how checks are passing though.

Also, I think requirements in `setup.cfg` and `setup.py` should be in sync.

edlerd:
Updated `setup.py` and refined the dependency definitions.
Successfully tested it locally with the anbox cloud dashboard in canonical/ams

edlerd:
> Updated `setup.py` and refined the dependency definitions. Successfully tested it locally with the anbox cloud dashboard in canonical/ams

setup.cfg
```diff
@@ -67,7 +69,8 @@ prometheus =
 	prometheus-client<0.8;python_version>"3.5"
 flask =
 	flask~=1.1;python_version~="3.5.0"
-	flask<3;python_version>"3.5"
+	flask<3;python_version>"3.5" and python_version<"3.9"
+	flask>=3,<4;python_version>="3.9"
```
maxiberta, line 73: This is gratuitously forcing users on recent python versions to use Flask 3. The Store is one such case for example :) running on Jammy (py310 ) on Flask 2.3.2. 

```sh
$ env/bin/pip install -e .[flask] flask==2.3.2
...
ERROR: Cannot install flask==2.3.2 and talisker[flask]==0.21.5 because these package versions have conflicting dependencies.

The conflict is caused by:
    The user requested flask==2.3.2
    talisker[flask] 0.21.5 depends on flask<4 and >=3
...
```

maxiberta: Can we maybe just `flask<=3` for any python version?

maxiberta (changes requested):
I think the updated requirements are too aggressive and will needlessly affect users - including our own services in production. Please make sure to keep full backward compatibility.

setup.cfg
edlerd: Sure, we can be more relaxed in the requirements here and let downstream projects choose the version they want. I'll adjust it here for flask and the other requirements.

requirements.docs.txt
```diff
@@ -1,2 +1,3 @@
 Sphinx==5.3.0;python_version=="3.6"
 Sphinx==7.0.1;python_version>"3.6"
+werkzeug
```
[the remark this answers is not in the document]
edlerd: The docs build in GitHub actions was failing without this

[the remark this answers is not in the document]
edlerd: Meaning the readthrtocs build, and think it does not run as GitHub action but in another external ci system.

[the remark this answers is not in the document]
edlerd: Here is the last failure, without this line: https://app.readthedocs.org/projects/talisker/builds/27508812/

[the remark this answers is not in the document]
maxiberta: Oh I see. Ok! Thanks

maxiberta (approved):
LGTM +1 thanks

PR #617, opened 2026-03-16
Fix building docs

jugmac00:


maxiberta (commented):
Thanks! I think it should be `setuptools<82` - [version 82 dropped `pkg_resources`](https://setuptools.pypa.io/en/latest/history.html#v82-0-0) which crashes `make docs`, in my local env at least.

jugmac00:
Thanks, @maxiberta, I pushed a new commit.

maxiberta (approved):
LGTM thanks!

PR #620, opened 2026-05-26
Drop project metadata from setup.cfg

maxiberta:
Fixes #619.


Perfect5th (approved):
Code changes seem entirely fine to me. Build and install works on my machine.

Tests have been broken for at least a couple of months – that's a separate issue to deal with.

jugmac00 (approved):
Thanks for looking into this!

Once we would be able to drop this one `setuptools==44.0.0;python_version<"3.10"` we could think about moving to pyproject.toml - but especially Launchpad is not yet there.

